PR #380, opened 2014-03-27
Test pipeline setup

hammer:
I've written a few unit tests here to exercise the pipeline creation code path. Note that there are some related data files that need to be added to the S3 bucket for these tests to pass. A few of these tests need some assert statements to be added, which I'll do in a subsequent commit. I just wanted to get the PR out there to get comments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ tests/unit/unit_test_data.tar.gz
 .#*
 .ropeproject
 .env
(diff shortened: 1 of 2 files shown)


roryk:
Hi Jeff, we moved away from running unit tests at all and instead run more integration type-tests with Travis. Sorry to close this after so much time. Thanks!

PR #704, opened 2014-12-18
upload raw collectl data to s3 after the run finishes

jwm:
Hey Brad, I think this will save collectl data in S3 after a run finishes, but I haven't had a chance to try it. Can you take a look and let me know if it looks good?


jwm:
Gonna table this for now given the open questions we talked about. We can always revisit at some point in the future.

PR #718, opened 2015-01-09
Add bsdmainutils package to install 'hexdump' 

tetron:
Required to build openssl package used by anaconda.


chapmanb:
Peter -- thanks so much for the report and the quick fix.

PR #753, opened 2015-02-10
Disabled DEXSeq when the aligner is bwa.

matted:
Newer versions of DEXSeq require the NH flag, which bwa doesn't set (see [here](http://seqanswers.com/forums/showthread.php?t=34990)).  This small workaround just turns off DEXSeq quantifications when bwa is the aligner so that the analysis doesn't crash.  This is probably a rare pipeline configuration (RNA-seq with bwa), but we do use it sometimes.

### DIFF
--- a/bcbio/rnaseq/dexseq.py
+++ b/bcbio/rnaseq/dexseq.py
@@ -46,6 +46,10 @@ def run_count(bam_file, dexseq_gff, stranded, out_file, data):
         logger.info("DEXseq is not installed, skipping exon-level counting.")
         return None
 
+    if dd.get_aligner(data) == "bwa":
+        logger.info("Can't use DEXSeq with bwa alignments, skipping exon-level counting.")
+        return None
+
     sort_flag = "name" if sort_order == "queryname" else "pos"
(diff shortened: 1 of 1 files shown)


roryk:
Matt, 

When are you doing RNA-seq with bwa? Thanks for your contributions lately, you are awesome. If you guys are rocking bcbio-nextgen a bunch, could you maybe add a blurb to https://github.com/chapmanb/bcbio-nextgen/blob/master/docs/contents/introduction.rst#users? It helps us to try to get funding.

matted:
We originally tried bwa with the RNA-seq "analysis:" pipeline as a testing setup when we were trying bcbio-nextgen for the first time.  Now we use it sometimes when we want read counts over genes but aren't worried about splicing much (mainly with yeast).  Also, it completes faster than STAR or tophat, which is nice.

Sure thing, I'll add something to the users file later today.

roryk:
Great, thanks a lot Matt. We've started doing some work with bacteria, and if you have any suggestions for improvements for handling the simpler organisms that would be sweet.

PR #754, opened 2015-02-11
Added paragraph about Gifford lab at MIT CSAIL.

matted:
Added a blurb about our lab to the users page.


roryk:
Hi Matt,

Thanks so much for taking the time to fix a bunch of issues lately and taking the time to write this up, we really appreciate it.

PR #755, opened 2015-02-11
-p option missing in a couple os mkdir instances

nuin:


chapmanb:
Paulo -- thanks for the fix. Much appreciated.

PR #827, opened 2015-04-14
Reflect new direct licensing of GATK & muTect

tfmorris:
The Broad Institute is switching to direct licensing on April 16 and Appistry will no longer be involved.


chapmanb:
Tom -- thanks much for being on top of this and for the documentation fix. Much appreciated.

tfmorris:
Thanks for the quick merge.  I didn't do anything special other than to click through the download/licensing links and notice the discrepancy.

PR #842, opened 2015-05-07
Add the paper published in Annals of Oncology in 2015 to the Mario Negri part of the User section

lbeltrame:
As the title says. In the interest of full disclosure, I'm the first author of said paper, and Brad is a co-author.


chapmanb:
Luca -- thanks so much for this. Congrats again on the paper.

PR #870, opened 2015-05-30
Minor workaround to avoid crash when no variants are found.

matted:
I ran into a problem running small test cases that didn't have enough reads to call variants, which crashed the pipeline at effects.py line 29:

    ann_vrn_file, stats_file = snpeff_effects(in_file, data)

`snp_effects` returns `None` in this case (`vcfutils.vcf_has_variants` returns `False`), which can't be unpacked to two variables.  This fix makes `snpeff_effects` always return a tuple.


chapmanb:
Matt -- thanks much for catching this. Sorry about the oversight and thanks again for the fix.

Issue #763, opened 2015-02-15
Logo ideas


robinandeer:
bcbio-nextgen is a cool project and branding is fun; let's all make some logos.

I'll go first:
![bcbio-nextgen-v2](https://cloud.githubusercontent.com/assets/1220232/6202538/9e3eb248-b4f1-11e4-85ec-51b59412b5f9.png)

[svg file](https://dl.dropboxusercontent.com/u/116686/bcbio-nextgen-v2.svg)
License: public domain / CC0

FYI: inspired by iojs/io.js#37

chapmanb:
Robin;
Thanks so much for starting this and the logo idea. From my side, I think we can drop the nextgen part in a logo and just use bcbio. Hopefully that'll make it a bit easier/cleaner with less text. Also the phrase nextgen is getting a bit anachronistic now; it was cooler 4+ years ago. Thanks again for getting this started and your logo skills.

lpantano:
nice!

i would put bcbio below the blue circle. agree with brad to use only bcbio.

thanks for starting this

roryk:
Robin, that's awesome!

robinandeer:
Thanks a lot everyone for both feedback and appreciation!

@chapmanb: I agree, putting "nextgen" on something is always tricky. Like video game consoles that are permanently labeled "nextgen", hehe.

New slightly updated logo:

![bcbio-nextgen-v3](https://cloud.githubusercontent.com/assets/1220232/6204704/093be9fa-b554-11e4-913b-74d0bee2bce0.png)

And by coincidence I noticed you've gotten over a fancy URL you might want to use for something in the future ^^

![bcbio-nextgen-v3-url](https://cloud.githubusercontent.com/assets/1220232/6204708/20b787b0-b554-11e4-9f4b-bf13b7eadaf8.png)

[svg file](https://dl.dropboxusercontent.com/u/116686/bcbio-nextgen-v3.svg)
License: public domain / CC0

guillermo-carrasco:
Oh! I really this last one that actually looks like a pipe :-D

Great work Robin!

lbeltrame:
Nice stuff!

kern3020:
Robin I love you work. This is the logo for bcbio in all its glory. While bcb.io is the url to the blog, would it confuse people with respect to pronunciation (ie, bcb - io) or wonder what bcb stands for?  I prefer the simpler bcbio.

roryk:
[...]

PR #878, opened 2015-06-04
add logo to README and docs

robinandeer:
I was just looking at the notes from yesteryear's code fest and noticed this:

> Future urgent task: need a LOGO

As per the consensus (I think :sweat_smile:) that was reached in #763 I now just went ahead and implemented the logo suggestion ^^

I think the banner looks quite good - please be critical though and I'll add some finished touches!


chapmanb:
Robin -- this is so awesome, thank you for doing it. You are really good at this.

One small practical question -- do we need to the README.html for anything in particular that the .rst doesn't cover? I don't mind having it but am always wary of keeping derived files that will eventually get out sync when I forget to update them.

Thank you again.

robinandeer:
Thanks Brad! I'm humbled :smile:

The ``.html`` wasn't supposed to be committed... I'll rebase it out of the PR

robinandeer:
Oops, didn't pay attention that you'd already merged - Get rid of the HTML file

PR #889, opened 2015-06-17
Refactor bcbio.rnaseq.rsem module

alexandrucoman:


chapmanb:
Thanks much for looking over this code and for the updates and fixes.

PR #892, opened 2015-06-17
Add support for Azure Blob storage service

alexandrucoman:
I have refactored the ```bcbio.distributed.objectstore``` in order to ease the future adding of one storage service.
For each storage service will be required one ```StorageManager``` and one ```FileHandle```.

The ```StorageManager``` is an interface which can ease the interaction with the storage service and is  exposing the following methods:
```

    def check_resource(self, resource):
        """Check if the received resource can be processed by
        the current storage manager.
        """
        pass

    def parse_remote(self, filename):
        """Parse a remote filename in order to obtain information
        related to received resource.
        """
        pass

    def connect(self, resource):
        """Return a connection object pointing to the endpoint
        associated to the received resource.
        """
        pass

    def download(self, filename, input_dir, dl_dir=None):
        """Download the resource from the storage."""
        pass

    def list(self, path):
        """Return a list containing the names of the entries in the directory
        given by path. The list is in arbitrary order.
        """
        pass

    def open(self, filename):
        """Provide a handle-like object for streaming."""
        pass
```

And the ```FileHandle``` is a handle-like object which facilitates interaction with the remote file.

This patch contains the ```StorageManager``` and the ```FileHandle``` for:
 - Amazon Simple Storage Service (Amazon S3)
 -  Azure Blob storage service

Thanks,
Alex

### DIFF
--- a/bcbio/distributed/objectstore.py
+++ b/bcbio/distributed/objectstore.py
@@ -7,10 +7,14 @@ Amazon Web Services S3.
 import abc
 import collections
 import os
+import re
 import subprocess
 import sys
+import time
 import zlib
 
+import azure
+from azure import storage as azure_storage
 import boto
 import six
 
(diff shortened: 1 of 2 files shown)


chapmanb:
Alex -- thanks much for this. We're hoping to roll out a new release in the next couple of days, and then we can pull this in and test for the following release. Much appreciated -- this is great work.

chapmanb:
Alex, thanks again for this and the other cloud generalization work.

PR #896, opened 2015-06-18
Refactor bcbio.graph.graph.generate_graph

alexandrucoman:
The files which contain information will be parsed in [bcbio.graph.resource_usage](https://github.com/alexandrucoman/bcbio-nextgen/blob/refactor/graph/bcbio/graph/graph.py#L250-L288) or another function and [generate_graph](https://github.com/alexandrucoman/bcbio-nextgen/blob/refactor/graph/bcbio/graph/graph.py#L291-L292) will receive [the ouput](https://github.com/alexandrucoman/bcbio-nextgen/blob/refactor/graph/bcbio/graph/graph.py#L365-L372) in order to plot the graphs.

This patch can be required by another cloud providers which have the resource usage information stored in another format.

Thanks,
Alex

### DIFF
--- a/bcbio/graph/graph.py
+++ b/bcbio/graph/graph.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 
+import collections
 from datetime import datetime
 import functools
 import os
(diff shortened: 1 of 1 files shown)


chapmanb:
Alex -- brilliant, thanks much for tackling this.

PR #901, opened 2015-06-24
Extended explanation of memory and cores

mpschr:
I propose the inclusion of the matters discussed in https://groups.google.com/forum/#!topic/biovalidation/R0-zqi55TMI in order for those things to be documented somewhere in the code.
Please correct the new sentences as needed (content, grammar).


chapmanb:
Michael -- thanks so much for this. It really helps to have users improving the documentation.

PR #921, opened 2015-07-02
Show batch name when paired batch misses paired samples (VarScan)

lbeltrame:
With large data sets, errors in batches (missing tumor/normal
phenotypes) can be confusing as it's hard to find the culprit. This
commit adds the batch name to the message printed, so it's easier to
spot errors.


chapmanb:
Luca -- thanks much for this improvement in the error reporting.

PR #929, opened 2015-07-12
Remove extra argument passed to FileHandle.__init__

alexandrucoman:
Fixes chapmanb/bcbio-nextgen-v#92


chapmanb:
Thanks so much for this quick fix.

PR #931, opened 2015-07-13
Made STAR handle Illumina 1.5 FASTQ properly.

matted:
As part of testing quality score logic, I realized that STAR pipelines don't convert quality scores to standard (Sanger / Illumina 1.9) when given illumina (Illumina 1.5) FASTQ input.  I made STAR use its quality score shifting parameter to fix this and get uniform bams.

### DIFF
--- a/bcbio/ngsalign/star.py
+++ b/bcbio/ngsalign/star.py
@@ -38,12 +38,13 @@ def align(fastq_file, pair_file, ref_file, names, align_dir, data):
     star_path = config_utils.get_program("STAR", config)
     fastq = " ".join([fastq_file, pair_file]) if pair_file else fastq_file
     num_cores = config["algorithm"].get("num_cores", 1)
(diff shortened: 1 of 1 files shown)


roryk:
Hi @matted,

Thanks for the pull request. I think downstream of the alignment there are some places where the quality flag gets used from the BAM file, and it figures it out by looking at the quality_format in the datadict so we'd have to fix that value too. It is a little confusing for us to swap the value in the middle of the pipeline but it makes sense to have everything put out the same quality scores.

roryk:
Thanks @matted, I think we're better off keeping the quality scores the way they are in the original files. I'm grateful you opened this though because if it turns out that is a bad idea we know where to look for the fix. I am willing to bet as time goes on it will be more of a bad idea because downstream tools won't handle the Illumina 1.5 quality scores right but right now it seems okay. Sorry to take so long to think about this.

PR #934, opened 2015-07-15
Relax lumpy exclude regions

mjafin:
to be able to detect  fusions in areas not covered by the capture bed file


chapmanb:
Thanks Miika

Issue #936, opened 2015-07-16
/bin/bash: {samtools}: command not found


funnell:
I'm getting these errors when running bcbio:
```
/bin/bash: {samtools}: command not found
...
cat: write error: Broken pipe
```

Which looks like it's due to this command (from bcbio-nextgen-command.log, scroll to end):
```
/share/data/apps/bcbio-nextgen/bin/novoalign -o SAM '@RG\tID:1\tPL:illumina\tPU:1_2015-07-07_test\tSM:C6736ANXX_5_1_chastity_passed' -d /share/data/apps/bcbio-nextgen/genomes/Hsapiens/GRCh37/novoalign/GRCh37 -f <(/share/data/apps/bcbio-nextgen/bin/grabix grab /share/scratch/tfunnell_temp/genome_instability/test/align_prep/C6736ANXX_5_1_chastity_passed.fastq.gz 1 20000000)    -c 16 -F STDFQ -r None -o FullNW -k | (echo '@HD   VN:1.3  SO:queryname' && cat) | {samtools} view -b - -o {out_file}
```

In bcbio/ngsalign/novoalign.py I've replaced a couple lines to look like this:
```
#cmd = cmd.format(**locals()) + tobam_cl
cmd = (cmd + tobam_cl).format(**locals())
```

Which seems to fix the problem:
```
/share/data/apps/bcbio-nextgen/bin/samtools view -b - -o /share/scratch/tfunnell_temp/genome_instability/test/align/C6736ANXX_5_1_chastity_passed/split/1_2015-07-07_test-sort-1_20000000.bam
```

If this looks okay I could submit a pull request.

roryk:
Looks good @funnell, thanks for the catch. Pull request away.

roryk:
Closing this since we fixed it. Thanks for the diagnosis!

PR #942, opened 2015-07-20
replace samtools placeholder with path in novoalign cmd

funnell:
addresses #936


chapmanb:
Thanks much for this and apologies about all the novoalign problems. I'm working now on ensuring all the different input options work cleanly with novoalign. Much appreciated.

Issue #975, opened 2015-08-14
Re-introducing parallel bgzip (pbgzip) during split align prep


schelhorn:
So I found the next bottleneck in our bcbio setup (there is an infinite sequence of these, so I'm not worrying too much about it). During sample/align preparation, bcbio converts `gzip`ed and BAM inputs into` bgzip`ed and `tabix`'d files for split alignment. However, the `gzip` to `bgzip` conversion is rather slow since it is running on only one core. Based on the [sources](https://github.com/chapmanb/bcbio-nextgen/blob/30c48f194a2031986b57cf210cc2e37fea9ae2c0/bcbio/pipeline/tools.py#L27), it seems as if bcbio once used a parallel `pbgzip` implementation. In our local installation, I reinstalled `pbgzip` from a [samtools fork](https://github.com/nh13/samtools/tree/master/pbgzip) and it performs just fine, scaling up to 8 cores until the `gzip` input stream becomes the limiting factor. I was wondering is the segfault issues reported in the sources are still valid, and if it is possible or desirable for the official bcbio installation to move back to `pbgzip` at some point in time.

chapmanb:
Sven-Eric;
It would be great if you'd like to investigate this. I'm agreed that bgzip is a big bottleneck and re-introducing parallel bgzip would be a nice improvement.

In my past experience, parallel bgzip would segfault on some samples without apparent cause. Or at least, I couldn't identify a single factor that caused it to fail in some cases and not in others, which didn't make it easy to produce reliable work/not work cases. I did do one pull request that seemed to improve things some:

https://github.com/nh13/samtools/pull/2

but it still failed after that and I kind of gave up as I don't really understand all of the magic in that branch. I was hoping someone would take this on and we'd have parallel bgzip but that didn't happen so we can re-examine.

I guess the next best step would be having a standard build of this and make it an option so we can start to find and categorize failure cases. Are you using the raw build from @nh13 implementation, or have adjusted at all? Thanks much.

nh13:
I would be happy to help here too. 

Thumb typed for added typos

schelhorn:
[...]

PR #982, opened 2015-08-25
Re-enable support for pbgzip

guillermo-carrasco:
In relation with #975 , this should enable the use of `pbgzip`


chapmanb:
Guillermo;
Thanks for this. Can we require an explicit `tools_on: [pbgzip]` flag in the `algorithm` section of runs to turn this functionality on until it's more stable? For example here's how we use that to turn on qualimap (https://github.com/chapmanb/bcbio-nextgen/blob/9e25468b87078161d81222233f33877f39c19ab4/bcbio/pipeline/qcsummary.py#L124). We'd need something along the lines of:
```
if "pbgzip" in tz.get_in(["algorithm", "tools_on"], config, [])
```
to the if statement. Thanks again.

guillermo-carrasco:
Brad;

Thanks for the tip, it makes sense. I just pushed this changes.

chapmanb:
Brilliant, thanks again.

PR #994, opened 2015-09-02
Save zipped reports for fastqc

guillermo-carrasco:
This should move the zipped reports to the resulting output directory. As a (desired) side effect it still removes it from `tx_fastqc_out ` as previously.


chapmanb:
Guillermo;
Thanks for taking a look at this. Does the code currently create zipped reports? I had thought it doesn't do this by default so we might need some parameter tweaking. We'd also need to transfer these into the final upload directory. If all works with just this small change happy to pull it but I'd imagined it would take more tweaks.

guillermo-carrasco:
Brad;

Sorry I didn't check this. However I think that the zipped reports are actually being created. The [code that calls `fastqc`](https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/pipeline/qcsummary.py#L441) calls it with the flags `-t`, `--extract`, `-o` and `-f`, which actually generates the zipped reports and extracts the file without removing it. i.e

```
$ fastqc -t 4 --extract -o fastqc test.fastq
...

$ ls fastqc/
test_fastqc  test_fastqc.html  test_fastqc.zip
```

Is that right?

About the final upload dir, I'm not sure I understand what you mean. Doesn't the `fastqc_out` directory where I'm moving the zip file end up under `wrok/qc/<sample_name>/fastqc`?

Thanks for your help

chapmanb:
Guillermo;
bcbio copies everything into a `final` output directory so you can remove the `work` directory on completion. By convention, it's typically `../final` relative to the work directory. We'd want to copy the zip file over there if folks will use it. The code that handles conversion to this directory:

https://github.com/chapmanb/bcbio-nextgen/tree/master/bcbio/upload

I don't really know what FastQC does right now. Did this work as you expect with the unit tests (say `./run_tests.sh devel`)?

guillermo-carrasco:
Thanks for the guidance Brad. 

So, after running the code (`run_tests.sh`) and debugging a little bit, I can confirm that `fastqc` is generating the zipped reports. Also, and that's a bit more unexpected, without further modification it ends up in the final upload directory:

```
[gcarrasco@node02 tests]$ ls test_automated_output/upload/Test1/qc/fastqc/
fastqc_data.txt     Per_base_N_content.tsv         Per_base_sequence_quality.tsv  Per_sequence_quality_scores.tsv  Sequence_Length_Distribution.tsv
fastqc_report.html  Per_base_sequence_content.tsv  Per_sequence_GC_content.tsv    Sequence_Duplication_Levels.tsv  Test1-sort_fastqc.zip
```
Note the zipped reports.

The logic I followed in the code is basically "put the zipped reports wherever the rest of the reports are copied", which in this case is `fastqc_out`. I imagine that `bcbio` is copying all that's in there to the upload directory.

chapmanb:
Guillermo -- thanks so much for testing. Glad this worked out with so few changes -- this is much appreciated.

PR #1007, opened 2015-09-08
"Local" cluster changes for graph/collectl (on bcbio-nextgen collectl part)

brainstorm:
Parses local nodes/hostnames from the bcbio logfile when running with "-c local" cluster flag/option.

Related to PR https://github.com/chapmanb/bcbio-nextgen-vm/pull/111


schelhorn:
Are there any plans to allow graphing resources also with vanilla Bcbio (bon-signatures) on a ipython cluster if the cluster nodes have collectl installed?

chapmanb:
Roman -- brilliant, thanks for all these fixes.

Sven-Eric -- we'd like to support it, although are a bit worried about the challenges of getting ti working reliably across a large number of clusters. Right now it requires ssh access to nodes. It also needs to have the ability to only pull collectl files in the timestamp region. Currently it grabs everything from the collectl logs which works okay for AWS where the cluster is up for the time of the run, but won't be practical for long running collectl processes. Beyond that we can see how well it works for people and define the conditions a cluster needs to meet to run it. Thanks much.

brainstorm:
Both @chapmanb and @schelhorn, really good points and feedback.

If we were to work further on this front, I would say that having a "phone home" feature to track execution/results of the pipeline would be super useful and interesting... perhaps also controversial as with the GATK:

http://gatkforums.broadinstitute.org/discussion/1250/what-is-phone-home-and-how-does-it-affect-me

@ohofmann

schelhorn:
Thanks for the very good points; I wasn't so much thinking about a true "phone home" than having a more transparent way to see what step the pipeline is in, and where the current bottlenecks are. Tracking the resource manager and the debug/command log can give you that, but it's not very convenient for production purposes where there isn't a need for detailed debugging. But it's of secondary importance, there are more important things to do such as CWL.

chapmanb:
Roman and Sven-Eric;
I'm against having a phone-home feature outside of locally populating some type of database that provides run information and query. I wouldn't want this to leak outside a user installation.

This would all be cool to have but does complicate the setup and administration on remote clusters. Practically I'm hoping bcbio can move to CWL then users can decide what underlying infrastructure they would want: something like Arvados that provides tracking and monitoring, or something more lightweight that is easier to install. Building this up right is a lot of work and it would be better to collaborate with someone doing this infrastructure work than try to do it ourselves.

brainstorm:
Thanks Brad, also my thoughts regarding infrastructure work/efforts to support analytics on bcbio, perhaps too much time would end up there. I thought about the phone-home as an **opt-in** (do no evil) instead of GATK default behavior... think launchpad's apport reporting system, for instance:

https://launchpad.net/apport

Would be useful to assist in debugging and/or fixing issues faster or even in advance, but it's also fine as it is now :+1:

brainstorm:
@schelhorn, I accidentally reached your wish/milestone ;)... I have a running ipython notebook server in the cluster and I can now examine/zoom the interactive plots (using `pandas-highcharts`) from collectl from Glasgow's cluster 2 :)

schelhorn:
Wow, that sounds nice! Any chance I could replicate that or is the setup too complex?

brainstorm:
@schelhorn you could merge my last commits from my branch:

https://github.com/brainstorm/bcbio-nextgen-vm/commits/master

Then open up the notebook: 

https://github.com/brainstorm/bcbio-nextgen-vm/blob/vardict_graph_steps/bcbiovm/graph/collectl_plots.ipynb

Quite beta stuff, handle with care ;)

PR #1020, opened 2015-09-16
Add logfile and timestamp timeframe limit when processing/pulling collectl raw files

brainstorm:
Now by default it pulls collectl files that fall within the range of the logfile. Otherwise it fetches an unnecessary amount of collectl rawfiles (GBs).


chapmanb:
Great stuff -- thanks so much.

PR #1027, opened 2015-09-23
Add .travis.yml file for travis integration

guillermo-carrasco:
This file tells Travis how to install and configure the pipeline, as well as how to run the tests.

Note though that the sole addition of this file is not enough, its also needed to:

1. Activate Travis for this repository in GitHub going to the repository settings in GitHub, Webhooks and Services, and activate Travis-CI
2. Activate the integration of this repository in [Travis-CI](https://travis-ci.org/) signing in with your account and activating `bcbio-nextgen`

Also, and this is important, since the Docker container is running the `bcbio-nextgen` code installed _inside_ the container, this is not testing the latest codebase. To do that, we should find a way of updating the code inside the container before running the testsuite. Is that something that could be easily done?

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+sudo: required
+
+language: python
+
+services:
+  - docker
+
+env:
+  - BCBIO_DOCKER_PRIVILEGED=True
+
+before_install:
+  # Temporal fix for networking problem: https://github.com/travis-ci/travis-ci/issues/1484
+  - echo "127.0.0.1 "`hostname` | sudo tee /etc/hosts
+  # Get and install anaconda for custom Python installation
+  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh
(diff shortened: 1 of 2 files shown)


chapmanb:
Guillermo;
Awesome, thanks so much for this. I'm pumped to have CI finally integrated. I think I've got everything connected for the next push to the repository. Fingers crossed.

For updating the code on testing, you can do by cloning the latest repository, then running `bcbio_vm.py devel setup_install`:

https://github.com/chapmanb/bcbio-nextgen-vm#updates

This might break if we require tool updates to match the code but that should be a good push to run more regular builds of the docker container, which we also need to be doing. Thanks again for this. So nice.

guillermo-carrasco:
You're welcome :)

Awesome, thanks! I assume that I can just add these commands to the travis file.

Also, I forgot to mention that you can, if you want to, [skip a build in travis](http://docs.travis-ci.com/user/customizing-the-build/#Skipping-a-build) by adding `[ci skip]` to the commit message. This is useful for updates like docs or so. 

Cheers!

PR #1045, opened 2015-10-01
Fix: Get hostname from local runs 

guillermo-carrasco:
Use the keyword "Timing" in the logs to determine if its a local or distributed run and generate the `hosts` dictionary accordingly. Ping @brainstorm .

Thanks @chapmanb and @ohofmann for the help.


brainstorm:
Great hack guille :+1:!

PR #1046, opened 2015-10-02
Package name typo fix in docs

travc:
`libperl-devel` doesn't exist on ubuntu (at least for 14.04), should be `libperl-dev`


chapmanb:
Travis -- thanks for catching this, much appreciated.

PR #1051, opened 2015-10-05
Update some ambiguous part of AWS docs

guillermo-carrasco:
Update some parts of AWS documentation that were, IMHO, slightly ambiguous, specifically the command to submit an analysis on a cluster with NFS FS, which was wrong, since was the same command to submit a local analysis.

### DIFF
--- a/docs/contents/cloud.rst
+++ b/docs/contents/cloud.rst
@@ -208,14 +208,12 @@ Create your project directory and link the global bcbio configuration file in th
 
     mkdir /encrypted/your-project
     cd !$ && mkdir work && cd work
-    ln -s ~install/bcbio-vm/data/galaxy/bcbio_system.yaml .
 
 - Lustre file system::
 
     sudo mkdir /scratch/cancer-dream-syn3-exome
     sudo chown ubuntu !$
     cd !$ && mkdir work && cd work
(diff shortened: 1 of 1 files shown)


chapmanb:
Guillermo;
Thanks much for this. It helps so much to have people going over the docs. All looks great except that you shouldn't need to symlink `bcbio_system.yaml` into the local directory. If it's not finding the default we should try to figure out what is wrong rather than having this workaround. Thanks again.

guillermo-carrasco:
You're right, the problem is that if you don't link the file, then you get an error along the lines:

> Couldn't locate global configuration file, including default location /home/ubuntu/install/bcbio-vm/galaxy/bcbio_system.yaml

The file is one level deeper in `/home/ubuntu/install/bcbio-vm/data/galaxy/bcbio_system.yaml`, so I guess the solution would be to move it to the correct default location, or modifying the actual default location.

What is actually the default location? Is it the installation directory?

chapmanb:
Guillermo;
Sounds like a bug and we can debug that separately. If you want to pull this in without the `ln -s` lines, we can do that and then tackle the bug in a separate thread, or work on it all here. Whatever is better for you. Thanks again.

guillermo-carrasco:
Let's do that, I just pushed a new commit removing the symlinking.

chapmanb:
Thanks so much.

PR #1053, opened 2015-10-05
Basic CWL no-container test

brainstorm:
According to docs, this simple test should be enough to run CWL tests with `--no-container` flag.


guillermo-carrasco:
Have you tried in the cluster? If running with `--no-container` you probably need the local installation of `bcbio-nextgen` (with tools and so on), which you don't have in Travis.

brainstorm:
Yes, @guillermo-carrasco, I just managed to make it work in the cluster after the last commits on PATH and HOME env vars and also tweaking cwl's directory `testinput-bcbio_system.yaml` file so that it points to tools installed in cluster paths:

```
  #picard: {dir: /usr/local/share/java/picard}
  picard: {dir: /cm/shared/apps/bcbio/20150720-devel/tools/share/java/picard}
  #rnaseqc: {dir: /usr/local/share/java/RNA-SeQC}
  rnaseqc: {dir: /cm/shared/apps/bcbio/20150720-devel/tools/share/java/RNA-SeQC}
```

Now I see where @chapmanb was aiming at with the `from_world` function in `bcbio.cwl.create.py`. I'll either make this test pass on travis with and/or without docker or expand the `from_world` function.

I think I'll focus on the docker CWL part passing on travis, what do you think, @chapmanb?

chapmanb:
It would be great to make it easier to run CWL tests but is probably too early to include this in travis since it doesn't even work with Docker locally. I'd add the tests, leave it out of Travis, then try to get a reasonable subset working locally before trying to push into Travis. Would this work?

PR #1054, opened 2015-10-06
Some docs amendments 

guillermo-carrasco:
While reading the docs, I found these minor things that could be changed.

Excellent software documentation, by the way :-)


chapmanb:
Awesome, thanks so much for all these fixes.

PR #1065, opened 2015-10-19
Vardict paired germline

mjafin:
Yo Brad,
This pull request will help us run paired VarDict bare, without any of the paired filtering. Should be used with caution.


chapmanb:
Thanks so much for tackling this Miika.

PR #1076, opened 2015-10-23
Use sample name for fastqc reports

guillermo-carrasco:


chapmanb:
Thanks so much Guillermo.

PR #1089, opened 2015-11-01
Added missing comma between rn5 & canFam3

mdeboth:


chapmanb:
Thanks much for catching this.

PR #1131, opened 2015-12-02
Fix the BlobHandle read file issue

alexandrucoman:
The current version of the BlobHandle doesn't act properly when
a user tries to read a file from Azure Blob chunk by chunk.

Thanks,
Alex


chapmanb:
Thanks Alex for being on top of this. Much appreciated.

PR #1166, opened 2015-12-17
changed the template for chip-seq analysis

mistrm82:
adapted this based on most recent run of bcbio chipseq


mistrm82:
Modified the text for ChIP-seq analysis in the config section

lpantano:
Thanks for that! improving docs is always good.

PR #1172, opened 2015-12-31
Cufflinks assembling should take library type into count.

dezzan:
Hello!

Just a small bugfix: cufflinks assemble does not currrently use library-type option. Unless used  the right option on the stranded data the following error occurs:

BAM record error: found spliced alignment without XS attribute


dezzan:
@chapmanb Brad, hi! Have a quick look, please.

roryk:
Thanks @dezzan, good catch!

PR #1184, opened 2016-01-14
Add serialize functionality example usage and plotting.

brainstorm:
Just a short explanation on how to extract the pandas dataframes from the pickled file.


chapmanb:
Roman -- thanks so much for the doc update.

PR #1201, opened 2016-01-25
Fix fastq downsampling.

vladsaveliev:
If the first element of rand_records is high, there is a chance that it will move to the end in the first iteration of the while loop, and write only 1 read. The rest N-1 readlines will return an empty line.


roryk:
Good catch, thanks for the fix Vlad.

PR #1202, opened 2016-01-26
Improve resources downloading on the Windows Azure infrastructure

alexcoman:
This pull requests tries to improve the features set available on the Windows Azure infrastructure.
Also, it contains some abstractizations of the client module ported from the develop branch of bcbio-nextgen-vm in order to avoid duplicated code.

Thanks,
Alex


chapmanb:
Alex;
Thanks so much for this. I've been going back and forth for the past few days about how best to integrate all this great work. Practically we've been trying to move the non-biology cloud interaction code out of bcbio-nextgen itself into bcbio-nextgen-vm. What we're working towards is having bcbio-nextgen itself be agnostic to the infrastructure used to run it. I guess I'm worried this adds to bcbio-nextgen functionality where we're looking to subtract from it to gain better abstraction long term.

Do you think this could fit in bcbio-nextgen-vm if we did some refactoring? If not, happy to merge as is and work on it going forward. I don't want to block your progress.

roryk:
Thanks, closing this as it is pretty old by now and discussion has stalled. Thank you!

PR #1204, opened 2016-01-27
Reverse order of xaxis and calculate deltas accordingly.

brainstorm:
Thanks @guillermo-carrasco. 

@lpantano: chime in if you have any hints regarding this piece of [matplotlib code](https://github.com/brainstorm/bcbio-nextgen/blob/master/bcbio/graph/graph.py#L166):

```
  top_axis.set_xticks([k for k, v in tick_kvs])
  top_axis.set_xticklabels([v for k, v in tick_kvs],
                             rotation=45, ha='left', size=16)
```

As it is now, the xaxis `steps` annotations get overlapped on the resulting graphs :/

skarmavbild 2016-01-27 kl 22 28 59

### DIFF
--- a/bcbio/graph/graph.py
+++ b/bcbio/graph/graph.py
@@ -99,7 +99,7 @@ def delta_from_prev(prev_values, tstamps, value):
 
     # Take the difference from the previous value and divide by the interval
     # since the previous sample, so we always return values in units/second.
-    return (value - prev_val) / (cur_tstamp - prev_tstamp).seconds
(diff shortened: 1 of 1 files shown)


chapmanb:
Thanks for these fixes so far, happy to get the correct ordering in. If anyone has any luck debugging the label problem that would be awesome as well. Much appreciated.

lpantano:
as told to @brainstorm, it seems this fixes it:

`mpl.dates.num2epoch(mpl.dates.date2num(tstamp)`

don't know what changed though. :tada:

PR #1206, opened 2016-02-01
Better looking collectl plots. Fix Y axis for network plots to easily compare.

brainstorm:
node13_net_bytes

@ohofmann, @lpantano


chapmanb:
Roman -- thanks for cleaning these up. The new plots look great.

PR #1214, opened 2016-02-04
Master

apastore:
I am trying to Fix the MuTect2 vcf to match other paired sample caller
and to support the gremline.py, but I have no idea how to deal with the
out_file


apastore:
I am trying to fix the MuTect2 vcf output to match oder caller and to work with downstream germline.py, but still failing...
I appreciate any suggestion!
Thanks!

chapmanb:
Alessandro;
Happy to have a look at this. I'd like to see if we can stream the output to avoid needing to write file intermediates. It's frustrating things like sample names don't get mapped to the output file, and I also don't see an obvious flag for tumor/germline which we might have to do from the tumor and normal LOD scores. Thanks for all the work so far.

apastore:
Thanks Brad,
I also noticed that some INFO filed necessary for the RTG validation are missing "INFO=VQSLOD", "GQ", "QUAL", "DP"

Eduardo-Alves:
mutect2 doesnt have REJECT, SOMATIC filters in the vcf file and the anaysis is crashing at Germiline Extraction with an error:
File "/usr/local/share/bcbio/anaconda/lib/python2.7/site-packages/bcbio/variation/germline.py", line 109, in is_germline
if "SOMATIC" in dict(rec.INFO):
File "cyvcf2/cyvcf2.pyx", line 1179, in cyvcf2.cyvcf2.INFO.__next_ (cyvcf2/cyvcf2.c:24884)
File "cyvcf2/cyvcf2.pyx", line 1126, in cyvcf2.cyvcf2.INFO._getval (cyvcf2/cyvcf2.c:23956)
KeyError

According to this post, all mutect2 output variants are somatic

http://gatkforums.broadinstitute.org/gatk/discussion/6768/what-does-mutect-output-means


(i also had to use mutect2 nightly because of a bug in mutect2 which doesn't handle -L parameter for tumor only. I am not sure if this bug affects all mutect2 results)
http://gatkforums.broadinstitute.org/gatk/discussion/6690/mutect2-tumor-only-mode-empty-vcfs

Thank you

Eduardo

example mutect2 vcf output (now both indels and snps are reported)

1 43815008 . TG T . clustered_events;homologous_mapping_event BaseQRankSum=-7.482;ClippingRankSum=0.401;DP=2026;ECNT=7;FS=3.007;HCNT=18;MAX_ED=80;MIN_ED=29;MQ=92.63;MQ0=0;MQRankSum=-6.091;NLOD=0.00;ReadPosRankSum=1.955;TLOD=38.77 GT:AD:AF 0/1:1989,30:0.015
1 43815009 rs121913615 G T . clustered_events;homologous_mapping_event BaseQRankSum=12.469;ClippingRankSum=2.206;DB;DP=2026;ECNT=7;FS=2.581;HCNT=18;MAX_ED=80;MIN_ED=29;MQ=92.63;MQ0=0;MQRankSum=-18.528;NLOD=0.00;ReadPosRankSum=1.570;TLOD=374.92 GT:AD:AF:PGT:PID 0/1:1888,131:0.065:0|1:43815009_G_T

chapmanb:
Alessandro and Eduardo;
Thanks again for this work and background. I finalized the MuTect2 integration and performed a validation against the DREAM synthetic 4 dataset (http://imgur.com/a/gqzwm). It looks quite good on this dataset, although this likely one used for tuning. This finalization is based on your patches above but I used streaming to avoid needing to write and manipulate temporary files. It now calls and finalizes cleanly, include germline extraction (although it doesn't really call germline) and validation.

I did not include converting AF to FREQ scores, as AF is also commonly used and I don't want to collect too many of these conversions into the caller integration where they're hard to tweak and extend. If we want to normalize VCF attributes, it would be a good idea to handle this as a separate tool outside of bcbio that we could use in pipelines for normalization. This is a great project if anyone wants to take it on.

Thank you again for helping with this integration. Happy to have it in bcbio for the next release.

apastore:
Thanks you for this Brad! 
Happy that now Mutect2 is properly integrated after your commit!

I was starting some work on Aberrant Splice. Do you think it would be interesting for Bcbio? Do you have some software that would be meaning full to integrate? I was checking MISO and rMATS

Thanks!

roryk:
Hi @apastore,

We've been moving more towards running Sailfish/Salmon/Kallisto and using Sleuth to call differential isoforms downstream. Some of the sleuth stuff is in `bcbio.rnaseq`. MISO is great and the idea behind using the insert size distribution as a prior is really good, but creating the indices is non-trivial and it doesn't handle biological replicates very well. I haven't looked at rMATS at all, so any input you have about it would be great. There are lots of software that will spit out a huge list of splice variants but they are mostly false, and we are trying to avoid perpetuating that problem. I feel pretty skeptical of anything that is calling de-novo junctions in terms of it not being a false positive generator, but I would be happy to be wrong.

roryk:
I have some code that is commented out in https://github.com/chapmanb/cloudbiolinux/blob/master/utils/prepare_tx_gff.py to make the MISO event database (search for make_miso_events). I remember it had some problems with some types of GTF but I forget not what the issues were.

apastore:
I was reading the miso docs https://miso.readthedocs.org/en/fastmiso/annotation.html. And I have generate the gff file with rnaseqlib <http://rnaseqlib.readthedocs.org/> as they suggest for GRCh37. would not be possible to provide them with the other rnaseq gtf file?

roryk:
Right, for GRCh37 and mm9 and what not it works fine but other non-Ensembl GTFs you have to do a bunch of massaging to make it work which is what the cloudbiolinux script did. I think the other downside of  MISO is you can't do anything more than compare two groups to each other, there is no way to model a model a mode complicated experiment.

PR #1242, opened 2016-02-25
Rmats

apastore:
this is a first tentative integration of rMATS (Multivariate Analysis
of Transcript Splicing).  http://rnaseq-mats.sourceforge.net


chapmanb:
Alessandro;
Sorry to be so slow in getting to these additions. We've been working on finalizing your MuTect2 work so we could look at this. If you still find this addition useful and can submit a clean pull request we'd be happy to work on it. I'd need advice from @roryk and @lpantano on how best to integrate this into the current RNA-seq workflow. Thanks again.

apastore:
Sure! I have realized that rMATS have some problem to work on trimmed reads bam files.  that’s way I have pull the commit out, not because of other reason! Do bcbio generate tophat2 (or star) alignment without trimming? 

Thanks!

PR #1244, opened 2016-02-25
Typo in Multiple files per sample parameters list

bcosc:


chapmanb:
Bryan -- thanks so much for the fix.

PR #1245, opened 2016-02-25
bioconda for private bcbio installation 

apastore:
The documentation report conda install -c bcbio bcbio-nextgen for a local bcbio install.
Update the local/private bcbio installation conda to bioconda in the docs.


chapmanb:
Thanks for catching this, much appreciated.

PR #1246, opened 2016-02-26
fix the  No such file or directory error

apastore:
install a local / private bcbio run in No such file or directory error (e.g.
 /bin/bash: bcbio/anaconda/envs/bioconda/bin/cutadapt: No such file or directory)

### DIFF
--- a/docs/contents/installation.rst
+++ b/docs/contents/installation.rst
@@ -309,6 +309,7 @@ with `source deactivate`. Activate the environment and install bcbio within it::
   git clone https://github.com/chapmanb/bcbio-nextgen.git
   cd bcbio-nextgen
   python setup.py install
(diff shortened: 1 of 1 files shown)


chapmanb:
Thanks for this. It's a bummer we need a manual symlink here but I can see the issue since bcbio uses the location of the anaconda directory to find some tools. @brainstorm @guillermo-carrasco -- I believe this is the approach you use to work with bcbio development, is that right? Do you have any workarounds/suggestions other than the symliks?

brainstorm:
Hi Brad! For development I think I just use `pip install -e .` with bcbio's cloned repository and with an active anaconda environment. Have not had the need to symlink binaries yet :-S

PR #1255, opened 2016-03-03
Master

apastore:
Hi Brad, Hi Roryl, Hi Lpantano,

This is a re-edit of the Chip-seq peak calling procedure to integrate
rMATS in bcbio. I have realized that rMATS at the moment do not support
trimmed align BAM, but only un-trimmed BAM


chapmanb:
Thanks for prepping this. Would you be able to try re-submitting this from a clean branch? Right now this has extra merges and also contains the MuTect2 changes which are no longer needed. Ideally we'd have a single commit or two with the rMATS changes. Thanks again.

lpantano:
As a another comment. Are you now running rMATs for paired samples? I think rMATs can handle replicates, that would be much better. As @roryk mentioned, we are worry rMATs give mainly false positive, so maybe it would be good to try to run some test data to know what we can offer. Probably an option would be to have it turn off for now, and then only activate it if people want to try. I ran rMATs recently with 20-30 mill reads samples and many replicates, and well, I wasn't so convinced by the top splicing sites, and took quite a long time.

PR #1256, opened 2016-03-03
rMATS

apastore:
Hi Brad, Hi Roryl, Hi Lpantano,
I Brad I hope now is better.

@lpantano I have set rMATS to run using replicate for both sample and
control.

in the phenotype you have to set control or a sample name and I have
add replicate that as to be set as an integer.

I definitely agree with your and @roryk comment on False positive issues


lpantano:
nice, always is good to run with all replicates. Can you change docs as well so people know how to set it up. Maybe some example yaml file is a good idea. :)

roryk:
Hi @apastore,

Sorry, I will get to this soon, I've been working on some of the RNA-seq stuff in bcbio-nextgen and need to figure out how this will fit in.

apastore:
Sure

roryk:
Heya-- I think we've mostly going in a different direction than rMATS, so sorry for not getting to this. Feel free to reopen if rMATS is something you all are using.

PR #1257, opened 2016-03-03
change docs for rMATS

apastore:
some info on how to configure rMATS on bcbio


roryk:
Thanks, sorry for not getting around to merging this, I think we've mostly gone in another direction from rMATS for bcbio.

PR #1389, opened 2016-05-13
Fixed a link in plink_to_vcf.py description

dbolser-ebi:


chapmanb:
Dan -- thanks much for catching these.

PR #1405, opened 2016-05-27
Make size optional for objectstore file-like objects

pansapiens:
This makes the API for objectstore file-like objects closer to the standard library file object.

It also fixes a bug when running:

`bcbio_vm.py template s3://bla.yaml s3://foo.csv`

which fails in my hands because the the S3 object is read without a size argument here: https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/workflow/template.py#L131

### DIFF
--- a/bcbio/distributed/objectstore.py
+++ b/bcbio/distributed/objectstore.py
@@ -87,7 +87,7 @@ class FileHandle(object):
         pass
 
     @abc.abstractmethod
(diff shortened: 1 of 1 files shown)


chapmanb:
Andrew -- thanks much for the fix. Please let us know if you run into any other issues at all.

PR #1444, opened 2016-06-24
Seq2C integration

vladsaveliev:


chapmanb:
Vlad -- thanks so much for continuing to push this forward. I'll look at ensuring we pass the full set of samples to the cohort level runs and we can start testing this. Great work.

PR #1457, opened 2016-07-06
Cleanups in the Tophat wrapper

cbrueffer:
- Remove unneeded closing()
- Utilize "with" in one more instamce
- Update a comment to reflect which TopHat versions are buggy

### DIFF
--- a/bcbio/ngsalign/tophat.py
+++ b/bcbio/ngsalign/tophat.py
@@ -5,7 +5,6 @@ http://tophat.cbcb.umd.edu
 import os
 import shutil
 import sys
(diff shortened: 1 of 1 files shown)


roryk:
Thanks for cleaning that up. I tested this out and it works fine-- I'm not sure why we weren't using that before. Maybe when we wrote the code the context closing wasn't implemented in pysam. Or maybe I'm just dumb. Could you squash these 3 commits together into one?

cbrueffer:
Done, thanks!

roryk:
Thank you so much!

PR #1468, opened 2016-07-13
Added gene_phenotype flag to VEP analysis

matthdsm:
Added gene_phenotype flag to VEP analysis, as this info is valuable in a diagnostics settings


chapmanb:
Thanks much for the VEP improvements. Much appreciated.

PR #1469, opened 2016-07-13
Seq2C annotation clean-up 

vladsaveliev:
Calling `add_genes` function with `max_distance=0`, so the call of `bedtools closest` will emulate `bedtools intersect`. 

Splitting annotated BED file ambiguous annotations, e.g.
```
chr1  123  343  GENE1,GENE2
```
becomes
```
chr1  123  343  GENE1
chr1  123  343  GENE2
```
as expected by Seq2C.

Also, fxied GTF to BED conversion off by one (GTF is 1-based, BED is 0-based).


chapmanb:
Awesome -- thanks for making this integration compatible. Glad to have this moving forward.

PR #1476, opened 2016-07-15
Further cleanups

cbrueffer:
- Use "with" to safely open files
- Use Pysam's API to check read flags


roryk:
Thanks again for these nice fixes, Christian. I checked our test data to make sure this was disambiguating the same reads and it is, so your clarity are all good. Thank you!

PR #1478, opened 2016-07-18
Fix various issues flagged by PyFlakes.

cbrueffer:
These are mostly accesses to undefined variables.  I have changed these
to what seems to make sense from my reading of the code, however I have
not tested them!

Given that the existing code cannot possibly work, these changes at least
shouldn't make things worse :-)


roryk:
Hi Christian,

Thanks so much for spotting these. Those fixes look right to me. You are awesome!

PR #1491, opened 2016-07-27
GEMINI supports only human genome build hg19/GRCh37 for now.

cbrueffer:
GEMINI only supports hg19/GRCh37 for now, don't try to build DBs for other assemblies.

### DIFF
--- a/bcbio/variation/population.py
+++ b/bcbio/variation/population.py
@@ -206,7 +206,8 @@ def _has_gemini(data):
             and os.path.exists(os.path.join(os.path.dirname(gemini_dir), "gemini-config.yaml")))
 
 def do_db_build(samples, need_bam=True, gresources=None):
-    """Confirm we should build a gemini database: need gemini + human samples + not in tool_skip.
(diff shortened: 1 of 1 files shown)


chapmanb:
Christian -- thanks for this. We're working on incorporating the new GEMINI approach that enables hg38 (and other species) support but great to have this check in place now. Thanks again.

cbrueffer:
Indeed, very much looking forward to that! Is there a work in progress branch somewhere yet? I'm currently looking into adding support for variant annotation with Annovar (since that's what we currently use internally), but I'd love to switch to GEMINI as soon as possible.

ohofmann:
No branch, but working with @roryk to make sure that Gemini "2.0" works as a drop-in replacement, then migrating annotation to vcfannot using the existing annotation sources before finally testing support for hg38.

cbrueffer:
Sounds sensible! Looks like I should have a closer look at vcfanno instead of trying to shoe-horn Annovar in.

PR #1500, opened 2016-08-05
Minor documentation improvements.

cbrueffer:
- Typo fix.
- Join lines to avoid unwanted vertical whitespace in the output.
- Mark up gvcf like the other options.


chapmanb:
Christian -- thanks for all these fixes and improvements. Much appreciated.

Issue #1511, opened 2016-08-11
TopHat wrapper uses mapped SAM format instead of BAM


cbrueffer:
Hi all,

not sure this is actually an issue; more of a question (if there is a better forum to ask code questions, please let me know; I couldn't find a gitter room for bcbio...)

bcbio/ngsalign/tophat.py calls TopHat with --no-convert-bam, which causes it to write the mapped file in SAM format. What's the reason for this? From the TopHat code it looks like --no-convert-bam actually converts the existing BAM file to SAM using "samtools view", so it's probably even slower aside from the space penalty.

roryk:
Hi Christian,

Thanks, so much, that is a great catch, I suspect that was leftover from some old functionality, where a version of Tophat was producing mangled read names in some cases and we had to fix them with sed.

cbrueffer:
That explains it, thanks for the change!

PR #1512, opened 2016-08-11
Remove unused list _out_fnames.

cbrueffer:
Small follow-up cleanup in relation to #1511 ; the (unused) list _out_fnames contained the now stale filename accepted_hits.sam.


roryk:
Christian, kills dead code dead. 

It made more sense when I was thinking in than when I typed it.

PR #1516, opened 2016-08-15
When running TopHat, use TopHat-Recondition to fix the unmapped.bam file.

cbrueffer:
Currently the "missing 0x8 bit" problem that unmapped.bam files have
is fixed using a home-grown solution.  Instead, use TopHat-Recondition,
which fixes this problem as well as two additional problem classes.

The issues are described in detail here:
Homepage:  https://github.com/cbrueffer/tophat-recondition
Publication   http://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-016-1058-x

Since this requires an external program, how to ensure it is installed (it is available through bioconda)? Is cloudbiolinux contrib/flavor/ngs_pipeline_minimal/packages-conda.yaml the right place?

### DIFF
--- a/bcbio/ngsalign/tophat.py
+++ b/bcbio/ngsalign/tophat.py
@@ -147,8 +147,9 @@ def tophat_align(fastq_file, pair_file, ref_file, out_base, align_dir, data,
                            ref_file, config)
     else:
         fixed = out_file
-    fixed = merge_unmapped(fixed, unmapped, config)
-    fixed = _fix_unmapped(fixed, config, names)
+    fixed_unmapped = _fix_unmapped(fixed, unmapped, data)
+    fixed = merge_unmapped(fixed, fixed_unmapped, config)
+    fixed = _add_rg(fixed, config, names)
     fixed = bam.sort(fixed, config)
     picard = broad.runner_from_path("picard", config)
     # set the contig order to match the reference file so GATK works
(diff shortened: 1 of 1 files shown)


roryk:
Hi Christian,

Thanks for the pull request, I'm not against calling out to another tool that does the job better. The tool would need to get installed via bioconda, I can help you make a recipe for that if one doesn't exist already. We'll also need to find the `tophat-recondition` program by looking it up, like             `tophat_recondition = config_utils.get_program("tophat-recondition", data)`, since it might exist in a couple different places.

cbrueffer:
Hi Rory,

a bioconda recipe already exists. Would ```contrib/flavor/ngs_pipeline_minimal/packages-conda.yaml ``` in ```cloudbiolinux``` be the right place to add it? I'll test the lookup line you suggested tomorrow and update the pull request. Thanks!

roryk:
Hi Christian,

Yup, that's the place. Thanks!

cbrueffer:
Updated per suggestion by @roryk

roryk:
Thanks Christian, this seems to work ok for me. The Tophat folks are also pointing people to use hisat2 instead of Tophat going forward, do you have much experience comparing the two? We've been delaying a little bit while hisat2 worked out some of its issues but things are looking more stable as of late.

cbrueffer:
Thanks for the merge! I just started looking at hisat2 (hence the recent index fixes), can't really give an opinion yet; we've used TopHat up until now.

PR #1521, opened 2016-08-18
Tophat fixes

cbrueffer:
More fixes for TopHat:
- actually use the merged BAM file (mapped + unmapped); previously the wrapper would continue with the mapped file
- fix transcriptome index file names for hg38 and hg38-noalt; previously the transcriptome indexes would be rebuilt during every TopHat run


roryk:
Looks great. Thanks!

cbrueffer:
Quick as usual, thanks Rory!

PR #1551, opened 2016-09-12
Fix NameError on Mutect2 missing batches error

lbeltrame:
The code was likely copy-pasted from somewhere else, but lacked
the "items" parameter, therefore any error there would raise a NameError
because of an undefined variable.

This patch simply adds the "items" parameter, and changes the user of
the function accordingly.


chapmanb:
Luca -- apologies about the likely confusing error message here and thank you for the fix.

PR #1557, opened 2016-09-13
return chroms in decreasing order of number of reads to aid in paralllelization.

brentp:
it can help to keep more cores busy if we process in order of number of reads. this probably has a minor effect but does improve my case a few percent when a relatively large chrom (hs37d5) occurs last in the bam. This reduces the amount of time when a single core (out of N) is left processing the last few large chroms.


brentp:
I'm closing this as it doesn't do much for the extra complexity.

PR #1588, opened 2016-10-04
Fixing Links

raonyguimaraes:
I corrected some of the links that were broken.


chapmanb:
Raony -- thanks so much for the fixes. Much appreciated.

PR #1617, opened 2016-10-31
Coverage QC updates

vladsaveliev:
Starting a pull request for changes that I've been working on. This is more like a suggestion PR, so feel free to let me know if you have a better idea regarding each point.

1. Fixing QualiMap running parameters:
 - Add `--skip-duplicated --skip-dup-mode 0` to report only marked duplicates.
 - Providing correct `-gd` parameter (HUMAN/MOUSE). It will make QualiMap plot the reference GC for comparison with sample GC content distributions. I also made a pull request to MultiQC QualiMap module to render the ref GC line (https://github.com/ewels/MultiQC/pull/325). This PR also has fixes for the cumulative coverage plot.
 - QualiMap's metric names are postfixed with "_1e7reads_est" regardless of full/downsampled mode. Fixed that as well.

2. MultiQC extracts the sample name for QualiMap from the BAM file name (for general stats) and from the base folder name (and for the plot data, e.g. `<sample>/raw_data_qualimapReport/insert_size_histogram.txt`). Currently it means that the MultiQC report for bcbio will contain samples named "<sample>-dedup" and "qualimap". I addressed this issue in bcbio: a symlink is created for the input BAM file named <sample>.bam and fed into QualiMap instead of <sample>-dedup.bam; and also an extra level of directory is created named after the sample (so instead of qualimap/*, QualiMap data is saved into qualimap/<sample>/*). This latter affected both RNAseq and DNAseq.
Also, search of secondary files for QC results was only 2 level deep, meaning that QualiMap's raw_data_qualimapReport content would never be parsed. I made a fix to walk all the way down the directory tree to find the secondary data. Let me know if there was a specific reason to search only 2 levels deep. Would it make sense to restrict it by 3 levels now?

3. Function "bcbio/variation/coverage.py:_median_coverage" in bcbio:
 - In fact calculates average coverage, not median;
 - If target is available, it miscalculates coverage. It uses samtools' idxstats which reports stats for full genome; but the denominator for average coverage is total bed size, not full genome.
 - Additionally, it makes sense to ignore flagged duplicates and unmapped reads.
I addressed those issues: it now uses different methods to find whole genome and target coverage, depending of BED file presence.

4. Function "calculate_offtarget":
Currently, the offtarget count is reported as the total properly mapped reads, minus properly mapped reads on target. I propose to ignore flagged duplicates as well. I also added a metric "usable reads" which represents percentage of mapped unique reads on target as a share to all raw reads. I renamed this function to "target_coverage_stats" and reporting the stats in _run_qc_tools/_run_coverage_qc.


vladsaveliev:
I also created a small PR for MultiQC_bcbio https://github.com/MultiQC/MultiQC_bcbio/pull/6

lpantano:
Hi,

I think this is great. 

I only saw one step where I have a comment. You added a new function: `run_sambamba_coverage_depth ` but we are already running this if we give a coverage file: `coverage` function in the same file. So maybe we need to decide to merge this functionality someway?  In this case, the target average coverage is parsed from the files from this function in `qc/multiqc.py` file.  So I see right now different values here: we have the total genome coverage, the target coverage when `variant_regions` is giving and the target coverage when `coverage` region is giving. 

Maybe we can discuss about what to do to normalize this step.

All this improvements are awesome. Thanks.

vladsaveliev:
Thanks a lot Lorena for your very quick feedback, and for bringing very good points.

So my idea is to report all "target"-related metrics and plots with respect to `coverage` regions. If they are not provided, then we are good to use `variant_regions` (except for the "Coverage Profile Along Regions" which is plotted only if "coverage" regions are specified). If none of them are in the config, then use the whole genome as the target.

In the code, coverage stats like avg depth and offtarget are calculated for separate 2 purposes: callable regions stuff, and coverage QC reporting. If no `coverage` is given, stats should be identical, so I tried to ensure they are not calculated twice. And I also tried to merge the functions that do the same stuff (e.g. encapsulate all those sambamba calls) and make the coverage code more reusable.

It is still might not be clear for the used what BED was used as "target" for the MultiQC report stats. I plan to add some general info about the BEDs above the table into the MultiQC report, like I've done for TargQC: 
screenshot 2016-11-01 08 07 24

chapmanb:
Vlad and Lorena -- thanks so much for this, happy to have this merged into the MultiQC report. We're hoping to have a release next week so please test away and let us know if you run into any issues. Thanks again.

PR #1630, opened 2016-11-05
Target coverage reporting updates

vladsaveliev:
Minor updates and clean ups for the code related to the target coverage reporting:

1. Added "on padded target pct" hidden metric. It's the share of reads that map on regions extended by 200 base pairs. We look at this metric in AZ to spot issues when reads get piled up in the ends of exons (padded on-target is expected to be 1-2% higher than the standard on-target, otherwise somethings is going wrong). I made it hidden by default, we'll make it visible in our internal reports.
2. Add information about used coverage and variant calling regions in MultiQC. It may be not clear for a person who looks at MultiQC what meant by "target" in all kinds of metrics, so I added a couple of lines about that.
3. MultiQC general read statistics: making all reported by sambamba for consistency and for keeping a reasonable column order in MultiQC. Some metrics like "mapped rate" are calculated multiple times (by "samtools stats", by qualimap, by bcbio), and the final numbers may vary slightly in different tools, so it's safe to calculate all of them by one tool for consistency. Also, this way we can control the order of columns and make the report a bit cleaner, which is nice.
4. Simplified over-engineered offtarget save-retrieve code, made some code cleaner.
5. Fix "dt = _fix_duplicated_rate(dt)" in multiqc.py

I tested it against different kinds of inputs, here are examples of reports with different cases of input `coverage`/`variant_regions`:

Both `variant_regions` and `coverage` are specified:
http://quast.bioinf.spbau.ru/static/bcbio/cov_and_vcr_different.html
cov_and_vcr_different

Only `variant_regions` are specified:
http://quast.bioinf.spbau.ru/static/bcbio/only_vcr.html
only_vcr

Both are specified, but with the same file:
http://quast.bioinf.spbau.ru/static/bcbio/cov_and_vcr_same.html
cov_and_vcr_same

Only `coverage` regions are specified:
http://quast.bioinf.spbau.ru/static/bcbio/only_cov.html
only_cov

Nothing is specified:
http://quast.bioinf.spbau.ru/static/bcbio/no_bed.html
no_bed


chapmanb:
Vlad -- this is awesome, thanks for all these improvements. It's so great to have better consistency across processes. The only thing I worry about here is the implicit caching in the sambamba module won't play well with CWL, where we need to explicitly define all of the files that are retained and passed. This makes an assumption of a shared filesystem so the cached files exist across processes, where in CWL it only brings in and uses explicitly defined files (like the offtarget_stats in the CWL defs). As we test this more with CWL in the next iteration we'll need ways to define and pass these cached files in a consistent way.

vladsaveliev:
Thanks Brad for merging the PR and explaining how caching works with CWL for me. It's now more clear for me why some stuff looks over-engineered. Am I correct that we need to make sure that all files that we checked with `utils.file_exists()` or `utils.file_uptodate()` anywhere in the code in order to reuse them, need to be declared in defs.py? And that's all that needs to be done to address CWL caching in CWL?

chapmanb:
Vlad;
In CWL every needed file is declared in defs.py and explicitly passed between functions. There is no assumption of a shared filesystem so it can work on cloud platforms where files get staged from places like S3. So that's right on, anything we depend on needs to be declared. I like the way you make things more transparent from the code side but we should also think of ways to define and retain these coverage cache files (or maybe convert to a single cache file that gets updated). I'm open to whatever works best from your perspective.

PR #1631, opened 2016-11-07
Improvements in caching bed files. Fix counting genes.

vladsaveliev:
Few clean-up and edge-case scenario updates before the release.

1. I reviewed caching of `clean_file()` and `add_genes()` across the pipeline and made sure that the proper prefix (like `cov-`, `svprioritize-`, `svregions-`) is provided wherever needed for correct caching. 
2. Counting genes for MultiQC: if the BED is pre-annotated, skip re-annotating and count original feature names. If we failed to annotate bed, skip reporting the count. Also handle comma-separated names and dots, and cleaning files before add_genes.
3. Reusing `clean_file()` in `seq2c.py`, cleaning redundancy in `_prep_bed()`.
4. Minor fixes and inline comments in `bedutils.py`.

### DIFF
--- a/bcbio/pipeline/qcsummary.py
+++ b/bcbio/pipeline/qcsummary.py
@@ -313,27 +313,27 @@ def _run_coverage_qc(bam_file, data, out_dir):
         out['Duplicates_pct'] = 100.0 * mapped_dups / mapped
 
         if dd.get_coverage(data):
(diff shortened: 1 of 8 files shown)


chapmanb:
Vlad -- thanks again for all the brilliant cleanup work. Great improvement.

PR #1642, opened 2016-11-14
Support for workdir on S3 

### LINKED ISSUE #1437: Enhancement proposal: Usage of S3 file system
Given that the bcbio [roadmap](https://github.com/chapmanb/bcbio-nextgen/blob/master/TODO.md) includes replacing the shared file system with S3 when running on AWS (which would be super-extra-awesome): there is a new fuse'd file system that enables mounting S3 buckets with local (non-root) user permissions that is high-performance and 'just works'. Obviously, some of the POSIX semantics is lost, but these aren't used by bcbio anyway if files are copied from local storage (permissions, directly appending to existing files). 

I tried all (well, about four or five) of the competing implementations, and this single `go` binary just blows them out of the water: https://github.com/kahing/goofys 

Perhaps this could enable a light/drop-in solution to running bcbio from S3 without changing too much of the underlying implementation?

tetianakh:
Resolves #1437.

Changes in bcbio:

* Refactoring of distributed/transaction.py: break up the code into smaller functions, add unit tests.
* Change default tmp dir location to /tmp/bcbiotx. We do not want to create a tmp dir in a working directory (unless specified otherwise in a config file).
* Attempt to unify the way file_transaction is used in the codebase. To make the behaviour of file_transaction dependent on configuration provided by user, we must always pass the config as an argument to the file_transaction. Otherwise, the behaviour specified in config _cannot be consistently enforced_. Only small number of file_transaction calls was changed.
* When moving a file from temp dir to a work dir, create an empty "flag" file, then check the sizes of the files are equal, and remove the "flag" file. In case of error, the flag file remains in the working directory. 
* Wrap certain steps of rnaseq and variant calling pipelines into file_transaction (see the changes in code).
* Minor bug fixes in multiqc, utils, alignprep
* Add integration tests for the pipelines which must work with a workdir on S3. To test if a pipeline works with a workdir on S3, one has to mount a bucket with goofys and change a work dir for tests. I've added a mount.sh script to tests, which should be run with `source path/to/mount.sh`. Sadly, running such tests cannot be a part of CI, because it requires access to and S3 bucket.

Installation of goofys fs and mounting of S3 bucket should be a part of provisioning, so it is not included in the bcbio.

Changes in test suite:
* Split tests into unit (do not require any additional dependencies, test only bcbio module) and integration (require tools and input data, perform disk IO, etc)
* Keep tests which depend bcbio_nextgen_vm separately
* Use py.test as a test running framework, because it is simpler, cleaner, outputs much more useful messages for failed tests, has a number of useful plugins, and takes full advantage of python's dynamicity. 
* Add test coverage to CI builds, separately for unit and integration tests.
* Run ALL tests in CI that do not require installation of additional tools (like GATK) or databases. Integration tests are run in the docker container, built on top of bcbio/bcbio using the Dockerfile_dev. 

WARNING!  To make some tests pass, I've had to install/upgrade the following dependencies:
multiqc-bcbio=0.1.6 
goleft=0.1.6 
tophat-recondition=1.4
As the list of dependencies is not a part of the codebase, I cannot include the requirement for those versions into the pull request. 

WARNING! Some tests are currently failing (see the CI builds for logs). Those tests are related to the parts of bcbio that I didn't have a chance to work with, and although I did my best to make them pass, I couldn't figure out how to do that in case of test_7_cancer from test_automated_analysis.py and test_5_find_fastq_pairs from test_pipeline.py.


chapmanb:
Tetiana;
This is so incredible, thanks for all the work that went into this. We pushed out the latest release and would be happy to get this merged. I think a couple of recent changes caused conflicts -- if you're able to get up to date with the latest master and resolve these we can get this in.

My major concern with the functionality changes is the decision to use a global temporary directory (`/tmp/bcbiotx`) as a default instead of local. I understand how this would be useful for your case, but for the currently more common shared filesystem case the other default is much more likely to work out of the box. `/tmp` is pretty likely to get filled up on most systems with big workflows, hence the default to a local disk unless specified otherwise. Could we swap this back to using the current default with an easy way to switch over to the global temporary directory?

For testing, would it also be possible to update the documentation for running the tests (http://bcbio-nextgen.readthedocs.io/en/latest/contents/testing.html#test-suite). I'm also sad to see `run_tests.sh` go as it avoiding folks needing to have the testing framework installed separately, although we could also add py.test as a top level dependency so it's present in the PATH alongside bcbio_nextgen.py. Did you have a specific reason for removing the small `./run_tests.sh` wrapper.

Thanks again for this, definitely happy to discuss and get this pushed as soon as we can so folks can begin using it.

schelhorn:
I agree, local temporary disks should be used wherever possible to prepare data for transfer to S3. @tetianakh, I believe that is also what we intended, isn't it?

tetianakh:
@chapmanb thank you very much for the feedback! Sorry for the ling reply, I was on a conference, but now I'll implement the fixes ASAP.

tetianakh:
@chapmanb why is it needed to pass a specific directory location for the temporary directory (the second argument for `tx_tmpdir`)?

chapmanb:
Tetiana -- thanks so much for looking at these. The second argument for `tx_tmpdir` is optional and not required if you pass the `data` object since it pulls the choice from the configuration or else defaults to the current directory. `base_dir` just provides a different way to use this function if you didn't have a `data` object or wanted to specify some different location, but can generally be left as None for the work you're doing. Hope this helps.

tetianakh:
@chapmanb 
I've brought back run_tests.sh (changing the test runner inside to py.test), changed the default tmpdir to current working directory, brought back base_dir argument into tx_tmpdir, and updated the docs. 

I advocate for keeping dev and testing dependencies separately from the rest of dependencies. Bcbio installs hundreds of packages -- 656 only via conda -- and throwing in something which is not required to run bcbio on production won't help in managing the dependencies. py.test and fiends can be installed in a virtual environment (or in a conda environment), eliminating the need to determine where the test runner is via `run_tests.sh`. The command to run tests is simply `py.test`,  or `py.test tests/integration` to run a subset of tests, so wrapping it in a script seems redundant in my humble opinion, that's why I removed it initially.

chapmanb:
Tetiana and Sven-Eric -- thank you again for this great contribution. I merged into development and we can iterate and test with this all included. I also added you to the list of contributors, please let me know if you need a different affiliation or anything else. We really appreciate this. I'm looking forward to experimenting with the S3 mounted support and learning all about the wonders of py.test.

chapmanb:
Tetiana;
I'm working on updating the development resources for the bcbio install so it gets everything we need for testing and have a quick question: what was the motivation for needing multiqc-bcbio, goleft and tophat-recondition with the standard bcbio install (as opposed to the tool installation)? Ideally we're only including python and other imported dependencies with bcbio and the tools are installed as a separate step.

tetianakh:
@chapmanb yes, they belong to tools, of course you're right.

PR #1658, opened 2016-11-28
Seq2C concurrency fix, and QualiMap BAM symlink

vladsaveliev:
1. Avoiding concurrency writing issue when preparing BED for Seq2C. I also moved `clean_inputs` into `sample.py` to avoid circular imports.

2. Small fix in QualiMap running. MultiQC reads the sample name as the base input BAM file name stored in genome_results.txt. Instead making a BAM symlink specifically for QualiMap (which led to uploading the whole BAM file to the final dir), make it sed the genome_results.txt to substitute the file name.


chapmanb:
Vlad -- brilliant, thanks so much for these fixes. I got them all merged in with the other upstream changes so should be good to go in development now. Much appreciated.

vladsaveliev:
I noticed the conflict and pushed another commit seconds after you merged. But Tetiana's fix to remove that symlink in qualimap.py works too, so not a big deal. Thank for merging Brad!

Issue #1653, opened 2016-11-25
using vep 86 in bcbio


matthdsm:
Hi Brad,
Is there a reason the latest version of VEP wasn't included into the latest release? I understand if it isn't possible to include all the modifications I've requested earlier, but I think as for the vep release, newer = better no?

Thanks again!
M

chapmanb:
We hadn't yet worked on any of the additional VEP modules you added, so haven't worked on bumping the database and tool version yet. I was hoping to get dbNSFP working and then see about the possibility of adding in the new modules. There is potentially some value in just increasing versions, but also downsides. It requires testing to ensure the new version is actually better, and also imposes the cost of re-building the data indexes for everyone. Definitely happy to support you if you have the ability to test the new version and include the additional modules, sorry we haven't had the opportunity to do so yet.

matthdsm:
Hi Brad,

Thanks for the quick response. What steps need to be taken for the newest release to be included? I saw that someone else had taken on the dbNSFP issue, as I haven't had much time anymore to further solve any issues. I do think this problem is resolved now?

As for the testing of the latest version, currently our pipeline is set up to just use bcbio for mapping, variant calling and QC. After this is finished, we use VEP v86 and all plugins mentioned earlier to annotate the resulting VCF, after which we use vcf2db to create a gemini compatible db (support for hg38). This method has yielded some great results, both for GRCh37 and hg38, with several clinical cases solved using the resulting db. So we know it actually works.

We'd be delighted if we could plug this method into bcbio, because of the underlying parallelization, which should speed up our analysis significantly. The recipe for vep 86 is in bioconda and works fine.
It also includes the files needed for the plugins to run.

I would like to propose the following steps to be taken
- [ ] Bump vep to v86
- [ ] Add ggd recipe for dbNSFP (hg38)
- [ ] Add ggd recipe for dbSCNV (GRCh37 and hg38)
- [ ] (if possible) Add mechanism that allows users to set the required dbNSFP fields manually (e.g. in the bcbio_system.yaml)
- [ ] (if possible) Add mechanism that allows users to activate/deactivate vep plugins.

I'd be happy to help develop/set this up. There is also still some work that needs to be done on the bioconda recipes for GeneSplicer and MaxEntScan, but perhaps I could ask some help on that front from the bioconda community.

I'd love to hear your thoughts on this. We're so close to being able to use bcbio as our go to pipeline for almost every routine analysis.

Thanks!
M

[...]

PR #1659, opened 2016-11-29
made dbNSFP fields modifiable, added dbNSFP for hg38

matthdsm:
Hi Brad

I've tried to make some modifications to how dbNSFP was called with VEP. As I said in #1653 , It'd be nice I the users could modify which fields are used.

This PR aims to do just that. However, I haven't been able to test this (how should I go about that?).
Do you think this seems allright?

this goes together with chapmanb/cloudbiolinux#233

Thanks
M

### DIFF
--- a/bcbio/variation/effects.py
+++ b/bcbio/variation/effects.py
@@ -131,6 +131,7 @@ def run_vep(in_file, data):
                 if is_human:
                     dbnsfp_args, dbnsfp_fields = _get_dbnsfp(data)
                     loftee_args, loftee_fields = _get_loftee(data)
+                    dbscsnv_args, dbscsnv_fields = _get_dbscsnv(data)
                     prediction_args = ["--sift", "b", "--polyphen", "b"]
                     prediction_fields = ["PolyPhen", "SIFT"]
                 else:
--- a/bcbio/variation/effects.py
+++ b/bcbio/variation/effects.py
@@ -190,6 +190,18 @@ def _get_loftee(data):
     args = ["--plugin", "LoF,human_ancestor_fa:%s" % ancestral_file]
     return args, annotations
 
+def _get_dbscsnv(data):
+    """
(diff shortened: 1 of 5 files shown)


chapmanb:
Thanks for this. The approach looks great to me. Could you make the default dbNSFP annotations the same as previously so that folks upgrading get the previous behavior? It's great to allow people to adjust but would like to make the experience consistent for anyone who is currently reliant on the existing behavior. Thanks again.

matthdsm:
Done!
Do you suppose this will work out of the box? I don't really have an idea how I should best test this before doing a PR.

matthdsm:
I tried adding the dbscSNV plugin in the latest commit, I thinks that should do the trick, but I might have forgotten to make some other changes somewhere. This does depend on the upgrade to vep_86, since the plugin file isn't included in the vep_83 recipe.

chapmanb:
Thanks for working more on this. I'm happy to merge the dbNSFP changes if you can separate out your branches again. No problem to close and re-open or whatever works best for you. For testing, you can install your development version of the code into your current installation (or a second development installation) and the details about how we typically are here:

http://bcbio-nextgen.readthedocs.io/en/latest/contents/code.html#development-infrastructure

Hope this helps.

matthdsm:
Ok, I think I reverted everything that doesn't belong here. If you don't see any more issues, merge away!

chapmanb:
Thanks, this looks great to me but I'm not able to cleanup the messy history of this because of the merges. Would it be possible to create a new clean branch submit a PR from that? You could do without any git magic with:
```
git checkout -b vep-dbnsfp
wget https://patch-diff.githubusercontent.com/raw/chapmanb/bcbio-nextgen/pull/1659.diff
patch -p1 < 1659.diff
```
and then submit from that new branch. I could do but wanted to give you credit for your work in the git history. I'll close this for now and can merge your new PR whenever ready.

PR #1665, opened 2016-11-30
made dbNSFP fields modifiable, added dbNSFP for hg38

matthdsm:
fixes botched commit history from #1659

### DIFF
--- a/bcbio/variation/effects.py
+++ b/bcbio/variation/effects.py
@@ -173,10 +173,9 @@ def _get_dbnsfp(data):
     https://groups.google.com/d/msg/gemini-variation/WeZ6C2YvfUA/mII9uum_pGoJ
     """
     dbnsfp_file = tz.get_in(("genome_resources", "variation", "dbnsfp"), data)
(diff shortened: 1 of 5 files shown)


matthdsm:
Thanks for the git magic pointer, otherwise it could've been a mess again!

chapmanb:
Perfect, thanks again for the work on this.

matthdsm:
Huraah! On to the next!

matthdsm:
Hi Brad, 

I just realized something. The modifications I made to make dbNSFP configurable only work if the're made before the installation..
So this will require some more I think. We should be able to have a configuration file somewhere post install, where users can modify the fields used.

M

chapmanb:
Matthias -- that's true, local changes would get over-written with new versions of the bcbio configuration files. We could migrate this to take the options from `bcbio_system.yaml` instead under a `resources` section:
```
resources:
  vep:
     dbnsfp_fields: your,custom,fields
```
and then bcbio would use the default otherwise. What do you think about that approach?

matthdsm:
Hi Brad, 

Looks very good to me! This also gives the users the option to modify the params for each run, so I wholly support this!

M

PR #1668, opened 2016-12-01
Copy all star output from tmp to work dir

### LINKED ISSUE #1667: Oncofuse broken due to transcactional fixes to STAR module
Hello @roryk and @tetianakh,

since your fixes to the transactional directories for STAR (thanks!), the file `Chimeric.out.junction` isn't copied over to the shared directory anymore; consequently, `oncofuse` fails since it cannot find that file. Could you please have a look?

tetianakh:
Copy all STAR output from tmpdir into workdir.
Changes:
 * move all STAR action under a single `file_transaction`. Previously a part that produces `final_out` was under a file_transaction, and the rest was under `tx_tmpdir`. The output that landed into a temporary directory created by `tx_tmpdir` got lost. 
 * change `out_prefix = os.path.join(align_dir, dd.get_lane(data), "star")` to  `out_prefix = os.path.join(align_dir, dd.get_lane(data))`, because that's what OncoFuse expects when looking for files

Resolves #1667 
@roryk I'd be grateful if you could do a sanity check of the fix. The main question is whether we can do `with file_transaction(data, align_dir) as tx_align_dir`. This can be done if the `align_dir` is a directory for STAR only, as it will remove whatever files there were before `file_transaction` was called.

### DIFF
--- a/bcbio/ngsalign/star.py
+++ b/bcbio/ngsalign/star.py
@@ -3,6 +3,7 @@ import sys
 import shutil
 import subprocess
 import contextlib
+from collections import namedtuple
 
 from bcbio.pipeline import config_utils
 from bcbio.distributed.transaction import file_transaction, tx_tmpdir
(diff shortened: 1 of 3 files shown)


roryk:
Hi Tetiana, sorry for the delay in getting to this, I missed this pull request and the tag somehow. It looks great; we shouldn't be using the STAR directory for anything else so swapping over to just using the alignment directory makes good sense.

PR #1673, opened 2016-12-06
source alternate dbsnfp fields from bcbio_system.yaml

matthdsm:
As discussed in #1665 

I think this should do the trick?

Thanks
M


chapmanb:
Matthias;
Brilliant, thank you. I pushed a small update to this to use a list of fields rather than a string, which lets us use standard YAML lists when specifying rather than needing a big long string (https://github.com/chapmanb/bcbio-nextgen/commit/078c2cfeb97a98b04aba8c97a32651e2ed852837). Thank you again for the work on this.

matthdsm:
Great! Thank you for the fix, it's better to adhere to the yaml standard indeed. Should I also update to docs? Or is that something you'd rather do?
M

PR #1677, opened 2016-12-07
document fastrna-seq

stephenturner:
see https://github.com/chapmanb/bcbio-nextgen/issues/1575#issuecomment-265565185


roryk:
Thanks Stephen for updating the docs!

PR #1678, opened 2016-12-07
add example fast-rnaseq template

stephenturner:


roryk:
Thanks for this, let me know if the fast RNA-seq part can be improved, not many folks are using it so feedback is helpful.

stephenturner:
Well, since you asked... 😄 

I think it'd be nice to be able to turn on things like FastQC or kraken... essentially to make it more modular, rather than all (`RNA-seq`) or next to nothing (`fastrna-seq`). Related to that, it seems the qualimap step in the regular RNA-seq pipeline is not multithreaded, and takes _forever_. Like, hours. Longer than anything else. An example of something I'd like to turn off on some occasions if it can't be multithreaded.

Another point of confusion for me, why sailfish in the RNA-seq pipeline but salmon in the fast pipeline? Why not salmon for both, or the ability to choose?

Finally, I don't think it's a big time sink, but it looks like it's rebuilding the sailfish or salmon index every time. Would it not make more sense to save the indexes in `genomes`?

roryk:
Could you add yourself as a contributor in the README? Thanks!

We're definitely swapping to Salmon, the kmer shredding algorithm in Sailfish destroys information which is retained by the MEM style quasialignments that Salmon does, and Salmon can spit out BAM files of the quasi alignments so we can theoretically do QC. I just haven't had the chance to vet some of the bias correction options yet so I don't want to swap what's there out from under folks since while I think Salmon is better, Sailfish does a great job.

roryk:
Thanks, Qualimap can be super slow; there are steps in there that are multithreaded but I think some of the tools aren't. @lpantano do you know where the bottleneck might be? Maybe we can do something to speed it up a bit.

The Salmon index is kept as a per-project thing because it isn't too much overhead to recreate the index and it lets folks run whatever versions of the transcripts they want to quantitate against; we update the pre-created tarballs pretty infrequently since we need all the downstream stuff like snpEff and VEP and whatnot to match up, so we're unfortunately usually a couple Ensembl annotation releases behind.

lpantano:
Hi,

Right now, I dont have a simple solution for that. Some times is fast, sometime is slow.  And I didn't find the culprit of that. 

There is an option that is not documented right now, because I guess is not fully tested, but if you add `qc` under `algorithm` section and give, for instance: `[fastqc, samtools]` it would do only those steps during QC.    

Right now, duplicates and rRNA are inside qualimap function, maybe we can add qualimap to things to turn off and separate rRNA and duplicates metrics from qualimap. what do you think @roryk ?

roryk:
Sure, it would be nice to know what step is going slow as well. Do the qualimap logs give some feedback about where the stuck steps might be?

stephenturner:
I believe the dexseq counting takes a while too. Would be nice to be able
to turn this off as well. Haven't used DEXSeq in years.

PR #1690, opened 2016-12-14
updated vep cache version

matthdsm:
To be merged after issues with recipe for v87 are resolved
bioconda/bioconda-recipes#3177

### DIFF
--- a/config/genomes/GRCh37-resources.yaml
+++ b/config/genomes/GRCh37-resources.yaml
@@ -1,4 +1,4 @@
-version: 29
+version: 30
 
 aliases:
   human: true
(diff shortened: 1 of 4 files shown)


chapmanb:
Thanks for this, happy to put this in place once the VEP issues get resolved in bioconda. For this patch, can you also increase the version numbers for each of the changed resource files? That ensures the new configuration gets pulled into place during a bcbio update. Thanks again.

matthdsm:
Done!

matthdsm:
Recipe is merged, so I think this can be merged too.

M

chapmanb:
Matthias -- thanks again for getting this all updated.

PR #1694, opened 2016-12-15
added genesplicer to config, updated docs for dbscSNV

matthdsm:
related to chapmanb/cloudbiolinux#238


chapmanb:
Thanks again for finalizing and documenting all this.

PR #1699, opened 2016-12-16
added dbscsnv installer option

matthdsm:


chapmanb:
Thanks for this, you're right on with what you need to include in the automated install as well. There is one more place to add it in addition to the two you identified:

https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/install.py#L727

Would it also be possible to prepare a clean pull request without all of the merges? These get confusing when looking back through commit history. You could start a new branch and just pull the single diff into it (https://patch-diff.githubusercontent.com/raw/chapmanb/bcbio-nextgen/pull/1699.diff). In the future, you might want to do `git pull --rebase` when trying to get up to date with a master branch since that will avoid all the extra "merge branch 'master'" messages. Thanks again for this.

PR #1701, opened 2016-12-16
added dbscsnv installer option

matthdsm:


matthdsm:
Done! Thanks for the tip about the rebase option, will use that in the future!
M

chapmanb:
Brilliant, thanks again for the additions.

PR #1707, opened 2016-12-21
Coverage QC: thresholds for completeness plot and other updates

vladsaveliev:
Region coverage QC (completeness plot) currently checks a questionable number of depth cutoffs (1x, 5x, 10x, 20x, 40x, ... 100x, - step of 10, but skipping 30x for some reason). I find the following set of thresholds more reasonable: 1, 5, 10, 20, 50, 100, 250, 500, 1000, 5000, 10000, 50000 (high thresholds are only exposed if there is coverage at least at one sample), that what's we are used to use in TargQC, but let me know if you have another opinion.

Additionally into the text coverage files I added an extra cutoff of 80% of average coverage, which is not going to go into MultiQC, but I'm going to use it to parse in my oncology reports, hope you don't mind.

Other minor changes in this PR include:
- fix of Duplicate pct metric for RNAseq,
- expose target/genome info in MultiQC only for variant calling pipelines,  
- some refactoring of average coverage calculation.

### DIFF
--- a/bcbio/bam/sambamba.py
+++ b/bcbio/bam/sambamba.py
@@ -66,21 +66,19 @@ def number_of_reads(data, bam_file, keep_dups=True):
     else:
         return _count_in_bam(data, bam_file, '', keep_dups)
 
-def number_of_mapped_reads(data, bam_file, keep_dups=True):
+def number_of_mapped_reads(data, bam_file, keep_dups=True, bed_file=None, target_name=None):
     # use idxstats if using an indexed lookup
-    if keep_dups:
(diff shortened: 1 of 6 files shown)


chapmanb:
Vlad -- thanks again for cleaning this up to be more useful for high depth targeted data as well. I'll work on testing this on WGS runs prior to the next release, and also will look into incorporating `goleft indexcov` (https://github.com/brentp/goleft/tree/master/indexcov) for areas where need estimates only and are scanning the BAM file.

PR #1711, opened 2016-12-23
Coverage QC optimisation for WGS

vladsaveliev:
Brad,

Following our email discussion, starting a PR for coverage QC optimisation in WGS. The first commit is to skip padded coverage QC calculation.  Will continue trying to remove redundant sambamba/samtools runs.

### DIFF
--- a/bcbio/bam/sambamba.py
+++ b/bcbio/bam/sambamba.py
@@ -37,7 +37,7 @@ def index(data, bam_fpath):
     return indexed_bam
 
 def work_dir(data):
-    return utils.safe_makedir(os.path.join(dd.get_work_dir(data), "coverage", dd.get_sample_name(data)))
+    return utils.safe_makedir(os.path.join(dd.get_work_dir(data), "coverage", dd.get_sample_name(data), "sambamba"))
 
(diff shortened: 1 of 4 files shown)


vladsaveliev:
About reusing the `samtools stats` results in `_run_coverage_qc`. I compared `stats` and `view` outputs more closely and figured out that `stats` skips secondary alignments for all metrics, and that is what causes the discrepancy. To address that I added `not secondary_alignment` as an extra filter to all sambamba calls that are expected to count mapped reads. I also added `not mate_is_unmapped` as I at looks like a reasonable requirement for coverage QC. 

So as a result, in `_run_coverage_qc` I got rid of one `sambamba.number_of_mapped_reads` call for unique reads. I also replaced the two sambamba calls in the very beginning that read from `bam.idxstats`, because `bam.idxstats` keeps secondary alignment.

I think I'm finished with this pull request (unless you have comments or suggestions).

Merry Christmas!

vladsaveliev:
I think now I'm finished with this pull request (unless you have comments or suggestions).

Merry Christmas!

chapmanb:
Brilliant, thanks so much for all the Christmas work. Hope you had a fun day and thanks for getting this all synchronized.

PR #1727, opened 2017-01-09
[WIP] iRods upload

matthdsm:
Hi Brad,

I've started adding the iRods upload option. I'm pretty sure the current code isn't functional, but I'd like to know if I'm going in the right direction with this.

Could you expand a bit on what the input variables `finfo` and `sample_info` contain exactly?

Thanks for the feedback!
Matthias


matthdsm:
@schelhorn, perhaps this can be of interest to you too?

M

schelhorn:
This is a great first step; however, I'd recommend going with the official python iRODS client for more functions and better compatibility: https://github.com/irods/python-irodsclient/blob/master/README.md

matthdsm:
Hi @schelhorn,

The python client was my first pick too, but that would imply the user has to leave his credentials in plain text in the config file to create the connection. From a safety point of view, I don't think that's advisable. Calling the CLI might be less efficient, but would depend on a preconfigured connection to an iRods zone, avoiding the use of credentials in configs.  An other option of course would be to only allow ticket based access.
Personally, I'm more inclined to prioritize safety and avoid any human readable credentials laying around.

The python client also isn't documented very well, so I might be missing some options. Your thoughts?

M

brainstorm:
From my past experience with iRODS, I'm with @matthdsm analysis. The python irods client wasn't reliable nor easily deployable and most pipelines resorted to the supported cmdline tools, unfortunately.

schelhorn:
Alright, go ahead then ;)

matthdsm:
Hi @chapmanb,

I was wondering of you'd have the time to scroll through the code en let me know if I'm missing something. I'm a bit at a loss on how to proceed from here.

Thanks!
M

chapmanb:
Matthias -- thanks for the work on this. I don't have any practical iRODS experience so am not sure how much I can help. Are there specific places you're stuck that I could help with?

For general questions, looking at the filesystem implementation (https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/upload/filesystem.py) is probably the best way to go to see how `finfo` gets translated into the filesystem directory structure. It splits by `sample` (sample specific files) and `run` (whole project files) and then copies them over. `sample_info` is the global `data` object and not used for filesystem copy since we don't need additional information from it. Hope this helps.

matthdsm:
Hi Brad, 

Unless you have any major objections, I think the upload module for iRODS is ready for testing. We'll have to see how it behaves in the real world. 
The only further work I think needs to be done is to the download part, but I don't really know how to begin there, so up untill now, there are only some placeholder functions present.

matthdsm:
I screwed up the git history in this one, so I started over in #1746, and I'll close this one. Sorry!

PR #1728, opened 2017-01-09
added missing comma to datatarget option list

matthdsm:


chapmanb:
Thanks much for catching this and the fix.

PR #1734, opened 2017-01-10
fixed wrong executable name for maxentscan

matthdsm:
maxentscan conda recipe installs maxentscan_score3.pl and maxentscan_score5.pl, so the executable "maxentscan" couldn't be found when called.


chapmanb:
Thanks much for the fix.

PR #1740, opened 2017-01-12
dbscSNV: fixed wrong variable name

matthdsm:


chapmanb:
Matthias -- thank you for catching this, much appreciated.

PR #1748, opened 2017-01-13
corrected VEP version

matthdsm:


chapmanb:
Thanks for the docs fix.

PR #1751, opened 2017-01-15
add remaining py3 compat

mr-c:


chapmanb:
Michael -- wow, thank you for this work. We're planning a new release very soon and I'll get this merged after that. I appreciate you looking at this. Python 3 support would be great, although there is going to be a lot of work ensuring all the bcbio python dependencies work with version 3. Internally we'll also need to start handling having both 2 and 3 present for external bioinformatics tools that only support version 2. Thanks for helping take the first step.

chapmanb:
Michael -- thank you again for working on this. It's great to start the process of moving to Python 3. Much appreciated.

PR #1752, opened 2017-01-15
Coverage QC fix for unpaired reads

vladsaveliev:
Addressing @sluke123's issue with missing "Avg_coverage" metric when running with single-only reads.

Also, slightly changed the stats validation in QC: if no mapped reads are found or no reads found at all in BAM, still reporting as much coverage metrics as possible. Downstream tools may depend on `project-summary.yaml` and it's good to promise that we report those metrics even in abnormal situations.


chapmanb:
Thank you Vlad, I'll work on testing this for the next release. This is the last update before we get that out, so appreciate you helping resolve it.

PR #1762, opened 2017-01-18
Variant QC upload fix

vladsaveliev:
After uploaded info `final`, both `variants/bcftools_stats/%s.txt` and `variants/snpeff/%s.txt` are merged into `qc/variants/%s.txt`. This PR makes them separate.
 
MultiQC SnpEff module reads the sample name from the file name, and the bcbftools_stats one reads it from the contents. So it's safe to rename the bcftools file and keep the snpeff file name. Other than MultiQC, not sure if anything other downstream needs to be addressed. Given that the files are merged and no one noticed before, probably it's safe?


chapmanb:
Vlad -- thanks so much for catching this and apologies about the mistake when I added in snpEff variant support. This is pretty new, and broken in the outputs, so makes sense to fix right away. Much appreciated.

PR #1770, opened 2017-01-20
cleaned up iRODS upload code, switched upload approach

matthdsm:
switch upload command from `iput` to `irsync`, which seems more dependable.
removed unused code and functions.


chapmanb:
Thanks much for the improvements and continuing to test on iRODS

Issue #1766, opened 2017-01-19
VEP still needs some fixes


matthdsm:
Hi Brad,

I've opened this issue because I've found some things that still bother me when using VEP. I'll try to fix most in an upcoming PR, but I'd like some comments from you/anyone interested. Feel free to strike through anything you're opposed to

- [x] add `dbscSNV` to the plugins used by default. Since `dbscsnv`has been added as datatarget, we can assume the user wants the annotation if it has been downloaded. Same principle as `dbnsfp`.
- [ ] the genesplicer plugin isn't included, even though the executable is installed and the option is included in the list of activated plugins.
- [x] ~~speaking of plugins. I feel we should move the plugin selection away from `bcbio_system.yaml` and into the analysis config. Does this config support nested options?~~
- [x] ~~there should also be an option to disable default options~~
- [x] ~~there should be a function that lists all arguments vep uses and removes the duplicates. it should perhaps also filter out options that are set using different arguments (eg `effects_transcripts`)~~
- [ ] deprecate the use of the `--fields` flag, since all this does is format the CSQ tag, without any added benefit.

I'm hoping to hear your thoughts on this. I'd like to get all the kinks worked out so we can clinically validate everything by May.

Thanks!
M

matthdsm:
Another question. Is there a reason arguments that contain a whitespace are added to the arguments list with 2 separate entries? Or is this just code styling? Sticking the 2 together woud greatly simplify the process of cleaning up the arguments list.

Thanks
M

chapmanb:
Matthias;
Thanks for continuing to work on the VEP integration. In general, I'd like to push back a bit on introducing a ton of configuration knobs for VEP. We try to avoid this approach in bcbio and focus on having a default approach that works for most cases with minimal tweaking where needed.

Is there any downside to running all of the plugins as long as the data is installed? If not, we should just pile everything on and let folks sort out what they need in downstream analysis. The temporary configuration in bcbio_system.yaml was more to test and develop this but it would be great to settle on more of a standard approach to take here so other people don't need to push all these buttons to get things to work.

If we have a add everything approach this changes a lot of your needs since we don't need to have functions to micromanage argument parsing and should make the code simpler. Do you think this would work? Thanks again for the discussion.

matthdsm:
Hi Brad, 
[...]

PR #1771, opened 2017-01-20
Fixes for VEP

matthdsm:
Hi Brad,

This PR goes with #1766 and should fix most of the issues.
I haven't put any work towards the `--fields` argument, seeing as I'd like to remove it completely and I don't really know your position on this.

I'll run some tests locally to make sure everything works like it should.
M


matthdsm:
Hi @chapmanb, 

I think this should be it as far as fixing VEP goes. 
The only remaining issues here are
- GeneSplicer still isn't included, although I hope this will be fixed be removing the trailing slashes
- The use of the `--fields ` argument hasn't been dropped, even though I think it should.

I also noticed the `dbNSFP` annotations don't seem to work anymore. The header and CSQ tag have all the info, but no variants were annotated (tested on a giab sample).

chapmanb:
Matthias -- thanks for all these improvements, happy to get these in. Have you been able to test the changes to the GeneSplicer directory to see if they work? Likewise, any insight into the dbNSFP annotation failure? It seems worth resolving these before making any other adjustments to the module. We typically use a SnpEff/GEMINI based annotation framework so I don't have a ton of experience here, so leaning on what is working in your workflows. Thanks again for testing and helping on this.

matthdsm:
Hi Brad, 
Unfortunatly, I haven't been able to test the modifications to Genesplicer (unless you can tell me where I can find the genome_resource config files post installation?) As for dbNSFP, this has always quite the annoying plugins, since it seems to be a bit hit and miss at times. I'll try rebuilding the dbNSFP file first and test it again.
Alternatively, perhaps we can collaborate with @brentp so vcfanno can handle the dbNSFP file, or even better, dbNSFP without the preprocessing.

M

matthdsm:
Hi Brad, 

your comment in my other PR gave me an idea. How about we put the extra plugins for vep under tools_on?
something like:
``` 
tools_on:
  - vcf2db_expand
  - vep_extras: [genesplicer, maxentscan]
```
although I'm not really sure `tools_on` supports nested options. What do you think? It fits the model of "tweaking defaults".

Thanks
M

chapmanb:
Matthias -- thanks for the continued work on this. Are there cases where you only want specifically one of genesplicer or maxentscan? If we could get away with a single `tools_on: vep_splicesite_annotations` that would help reduce the number of parameters. Sorry for the pushback but trying to fight back too many configuration parameters is one of the hardest things to do. Thanks again.

matthdsm:
Hi Brad, 

No problem, I'll merge the two options! If you find the time, could you have a look at the genesplicer issue? I've tried identifying the problem, but to avail. the resource doesn't get included into the config dict.

thanks!
M

matthdsm:
New PR with clean history #1784

PR #1772, opened 2017-01-20
Fix vcfanno dbSNP race condition when writing the config file

vladsaveliev:


chapmanb:
Thanks so much for identifying the issue and for the fix.

Issue #1775, opened 2017-01-23
vcf2db error messages


matthdsm:
Hi Brad,

While testing the latest development version I noticed vcf2db throws the following errors:
```
[2017-01-23T10:58Z] GEMINI: create database with vcf2db
[2017-01-23T10:59Z] skipping 'AC' because it has Number=A
[2017-01-23T10:59Z] skipping 'AF' because it has Number=A
[2017-01-23T10:59Z] skipping 'MLEAC' because it has Number=A
[2017-01-23T10:59Z] skipping 'MLEAF' because it has Number=A
...
```

I've encountered this issue before (quinlan-lab/vcf2db#14) and this can easily be solved with a small `sed` replacement > `sed -e 's/Number=A/Number=1/g'`

 I've found several places in the code where this could be added, but I don't know where you'd prefer it. Currently I think the best place to add this would be [here](https://github.com/chapmanb/bcbio-nextgen/blob/9d5232684763db34accb6837712bfcf9423c0a02/bcbio/variation/multiallelic.py#L67), right after decomposition.

Your thoughts? If this step isn't taken, all of the above columns would be missing from the db.

Thanks
M

chapmanb:
Fixed via #1777

PR #1777, opened 2017-01-23
fix for vcf2db warning messages

matthdsm:
see #1775


chapmanb:
Matthias -- thanks for identifying the problem and the fix. Much appreciated.

Issue #1573, opened 2016-09-26
HG38 and Gemini


matthdsm:
Hello again,

I was wondering if there are plans to overhaul the approach to gemini to support Hg38? According to [this thread](https://github.com/arq5x/gemini/issues/781) it's possible to use gemini with Hg38 using vcfanno and vcf2db. Do you think it is viable to replace the`gemini load` step with annotation by vcfanno and db loading with vcf2db? I suppose, this way it's possible to keep supporting hg19  and also build in some extra hg38 support.

M

brentp:
Hi Matt, I'm not clear on what you are asking. As I stated in the other thread, you can load with vcf2db and that will be the way forward. To my knowledge the loader is stable and integrates with gemini well. We just haven't aggregated the annotations yet, but users can do that on their own and send to vcfanno.

brentp:
oh, I just noticed this issue is in bcbio-nextgen. I just got a notice and assumed it was in gemini. Once we have everything in place, I will work to get it integrated into bcbio-nextgen.

We have a grad student working on aggregating annotations and developing the infrastructure for that now.

ohofmann:
@roryk is working on this within bcbio. Some more testing to do, and I want to run a couple of GiaB runs and ICGC samples (exome, WGS) to get a sense of performance and edge cases. Should be integrated soon (tm).

roryk:
It's on my TODO list but totally happy to have Brent work on it since he knows way more than me. Right now my progress was "the gem.conf in vcfanno seems to work ok on my test variant file".

ohofmann:
@brentp Are you going to focus on somatic use cases at all? If not that's something we might be able to field...

brentp:
no. we don't have any plans on somatic stuff. I am happy to help anywhere, as needed.

@roryk, I'll make a few updates to the gem.conf this week as we've found a few things that need tweaking.

brentp:
I have just updated the gem.conf and custom.lua within the vcfanno repo.

matthdsm:
Hi Brad,
I'd like to follow up on this question. Can Bcbio already create db's for analysis run on hg38? Or is that something that's still in the works? As it stands now, we let bcbio skip the gemini step and we then manually load the annotated vcf with vcf2db (which works great, parses the complete CSQ tag from vep). It would be great to have vcf2db integrated as an alternative to Gemini on hg38.

Thanks
M

chapmanb:
Matthias;
[...]

PR #1782, opened 2017-01-24
added expand flag to vcf2db command

matthdsm:
as discussed in #1573, added `--expand` flags to vcf2db command

M

### DIFF
--- a/bcbio/variation/population.py
+++ b/bcbio/variation/population.py
@@ -51,7 +51,8 @@ def create_gemini_db(gemini_vcf, data, gemini_db=None, ped_file=None):
         ann_file = vcfanno.run_vcfanno(gemini_vcf, "gemini", data, data_basepath)
         with file_transaction(data, gemini_db) as tx_gemini_db:
             vcf2db = config_utils.get_program("vcf2db.py", data)
-            cmd = [vcf2db, ann_file, ped_file, tx_gemini_db]
+            vcf2db_args= ["--expand gt_types", "--expand gt_ref_depths", "--expand gt_alt_depths"]
+            cmd = [vcf2db, ann_file, ped_file, tx_gemini_db] + vcf2db_args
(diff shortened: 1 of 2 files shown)


chapmanb:
Matthias;
Thanks for this. We'll probably need to make expanding these columns a configuration option since it impacts the speed and size of the output database.

The best approach might be to have an `tools_on` option for this (http://bcbio-nextgen.readthedocs.io/en/latest/contents/configuration.html#tweaking-defaults) so you can do:
```
tools_on: [vcfanno_expand]
```
that will add these options. What do you think?

matthdsm:
Hi Brad, 

Sound great to me, although I'd call the option 'expand_genotypes', to make it sound more generic. On that same note, I probably like to see the 'gemini' param changed to 'variant_db' or something, since strictly we aren't handling gemini db's anymore. 

Thanks!
M

chapmanb:
Mathias -- I prefer the more specific naming since the general ones can be misinterpreted. For example, if you just see `tools_on: [expand_genotypes]` it's very hard to guess what step that applies to and what is does. Similarly, `variant_db` is pretty generic and doesn't reflect that we're creating them in a certain way for querying with the GEMINI framework.

Probably `vcf2db_expand` is the right flag since that's actually the step this happens at, but happy to discuss more if you or others feel strongly about more generic names.

matthdsm:
Hi Brad, 

Frankly, the naming doensn't matter all that much to me. I mostly care about the functionality :) However, you're right the users should be able to know which step is affected by each parameter.
I'll look into adding the option `vcf2db_expand` in the next commit. After that, I suppose this can be merged.

Thanks for the comments!
M

matthdsm:
There we go, fixed and tested! If there aren't any new issues, I believe this can be merged.

M

chapmanb:
Matthias -- thanks so much for this.

PR #1783, opened 2017-01-24
First successful ubuntu 16.10 aws ansible run

brainstorm:
Solves ansible bootstrapping problems on new Ubuntu and unix socket issues:

https://github.com/ansible/ansible/issues/11536#issuecomment-120092814
https://stackoverflow.com/questions/32429259/ansible-fails-with-bin-sh-1-usr-bin-python-not-found#34402816


chapmanb:
Roman -- thanks so much for all these. Will be updating soon with better docs and more automation of VPC setup.

PR #1788, opened 2017-01-26
VarDict support: failing bcftools when no variants called

cbrueffer:
In case VarDict hasn't called any variants, the bcftools call added to the commandline in commit 83ff6cce76ad772f553be65f05efcd60ef7c7e9b results in an uncaught error.

With bcftools 1.3, this is the error message: "Failed to open -: unknown file type"

The patch in this PR handles the problem, but spams the log with stack traces. Calling do.run() with log_error=False would work, but potentially hide other failures.

There must be a better way, any suggestions?

### DIFF
--- a/bcbio/variation/vardict.py
+++ b/bcbio/variation/vardict.py
@@ -145,13 +145,20 @@ def _run_vardict_caller(align_bams, items, ref_file, assoc_files,
                             vcfutils.write_empty_vcf(tx_tmp_file, config, samples=[sample])
                         else:
                             cmd += " > {tx_tmp_file}"
-                            do.run(cmd.format(**locals()), "Genotyping with VarDict: Inference", {})
+                            try:
+                                do.run(cmd.format(**locals()), "Genotyping with VarDict: Inference", {})
+                            except:
+                                vcfutils.write_empty_vcf(tx_tmp_file, config, samples=[sample])
+
                 else:
                     if not _is_bed_file(target):
                         vcfutils.write_empty_vcf(tx_out_file, config, samples=[sample])
                     else:
                         cmd += " > {tx_out_file}"
-                        do.run(cmd.format(**locals()), "Genotyping with VarDict: Inference", {})
+                        try:
+                            do.run(cmd.format(**locals()), "Genotyping with VarDict: Inference", {})
+                        except:
+                            vcfutils.write_empty_vcf(tx_out_file, config, samples=[sample])
             if num_bams > 1:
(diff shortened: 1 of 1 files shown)


chapmanb:
Christian -- sorry about the issue. What version of bcbio are you seeing this problem with? There was a bug in latest development for a day (https://github.com/chapmanb/bcbio-nextgen/commit/9d5232684763db34accb6837712bfcf9423c0a02 to https://github.com/chapmanb/bcbio-nextgen/commit/3b1945f01064314611d6e2cd0947b5d20a5c62cb) 3 days ago that caused this problem by ending up with empty BED files feeding into VarDict. VarDict will produce a valid VCF header if there are no variants but wasn't doing the right thing with no BED files.

If you were running that version: updating, removing the vardict directory and re-running will fix. Hope this explains the issue and sorry about the problem.

cbrueffer:
Hi Brad, thanks for the feedback! This is with bcbio-nextgen 1.0.1 and vardict-java 1.4.8. When I first saw the error it was with a non-empty BED file, although I can't tell whether that's always the case.

cbrueffer:
This is the exact call that results in no output from VarDict-java; the BED file is non-empty:

vardict-java -G /casa3/nobackup/med-csb/software/bcbio/genomes/Hsapiens/hg38-mut-cb/seq/hg38-mut-cb.fa -f 0.02 -N XXXXXXX_1_l_r_m_c_lib_g -b /tmp/med-csb-runtest/SCANB-XXXXXXX/work/align/XXXXXXX_1_l_r_m_c_lib_g/1_2017-01-26_SCANB-XXXXXXX.bam -c 1 -S 2 -E 3 -g 4 -Q 10 /tmp/med-csb-runtest/SCANB-XXXXXXX/work/vardict/chr11_KI270721v1_random/XXXXXXX_1_l_r_m_c_lib_g-chr11_KI270721v1_random_0_100000-unmerged-regions-regionlimit.bed

cbrueffer:
var2vcf_valid.pl only prints a VCF header if it found any variants in its input. Modifying var2vcf_valid.pl to print the header unconditionally fixes the bcftools issue as well. Maybe we should add an option for this?

chapmanb:
Christian;
Thanks for the example case, this helps a lot. I tracked this down to a recent change in the var2vcf scripts only for non-paired cases. This got triggered when we updated the VarDict perl scripts for another case, apologies for missing it. I'll work on preparing a new bioconda package with the fixes and then will report back here. Thanks again for the help debugging.

cbrueffer:
Excellent, thanks Brad!

PR #1790, opened 2017-01-27
Fix race condition in coverage QC 

vladsaveliev:
In coverage bed cleaning and merging, and also padded bed preparation.


chapmanb:
Thanks so much for catching this and the quick fix, Vlad.

PR #1806, opened 2017-02-08
VEP: fix for genesplicer plugin

matthdsm:
Hi Brad,

Apparently, the genesplicer function called the executable dir, instead of the executable itself. An oversight which didn't come out during my testing, since the genesplicer resource seemed unavailable.

Anyways, this should fix it.

Thanks
M


chapmanb:
Awesome, glad this is working now thanks for testing it out. Much appreciated.

Issue #1808, opened 2017-02-08
How to accumulate QC files over multiple bcbio runs?


NeillGibson:
Hi Brad, 

We want to accumulate the QC files of the increasing cohorts of samples that we sequence. 

We re-variant call all samples when our cohorts expands (we get new samples)  to get a square genotype table over all samples.  
This leads to fastqc and samtools stats being run for samples for which we already have fastq and samtools stats files. See https://github.com/chapmanb/bcbio-nextgen/issues/1807

When a (re)variant calling is finished we would like to copy the samtools and fastq zip files to another directory were we accumulate them so we later can generate on-demand (aggregated) reports on those files with multi-qc or other tools. 

The only way we current see to do this is to do a find exec on the stats files in the bcbio final directory to copy them to the accumulation directory. E.g.
```
find /bcbio_run_123/final/ -name *zip -exec cp {} /accumulatated_QC_files/fastq \; 
find /bcbio_run_123/final/ -name *txt -exec cp {} /accumulatated_QC_files/samtools \; 
```
This would be easier if the samtools and fastq file names include not only the sample name but also `samtools` and  `fastq` in the name. These find execs currently select not just the intended QC files but also other files. 

Another option would be that bcbio outputs a list of QC files per sample. I am not sure how easy / difficult it would be to implement that but it sure would be useful for accumulating the QC files over multiple bcbio analysis runs. 

Thank you. 

Neill

chapmanb:
Neill;
Thanks for all this discussion, definitely re-running/combining from multiple projects is not a strong point of bcbio and happy to try and improve this.

The simplest way to do this right now is to save the sample qc directories from `work/qc/sample-name`. These are equivalent to the directories in `final/sample-name/qc` but named in an easier way for dropping into a new run. Putting them into `work/qc` before running will skip re-calculating them and include them in the final MultiQC report.

Unfortunately we don't have much flexibility on file naming in many cases since MultiQC needs to use the filename to determine the sample. So if we call a file `samplename-samtools.txt` it will call is `samplename-samtools` instead of `samplename`. This is true for any files where the same name is not embedded in the output report somewhere. So keeping the directory structure is the easiest way to approach it.

There is also #1708 which aims to something similar and make it easier to run a consolidated MultiQC following the initial run. This would be helpful in parallel to explicitly saving the work directories in re-runs, so you can re-combine with MultiQC without a re-run.

Does saving/copying the QC directories work for what you'd need?

[...]

PR #1821, opened 2017-02-18
MultiQC: upload file list into final

vladsaveliev:
1. Refactored the upload code a little to separate the functions that only return paths from the functions that actually upload files (e.g. `update_file` vs. `get_upload_path`)
2. Added a function that returns all paths that would to be uploaded for current run.
3. Use that function to generate `list_files_final.txt` for multiqc (partly resolves https://github.com/chapmanb/bcbio-nextgen/issues/1708)

Unfortunately that works only for the filesystem now. I don't have an infrastructure to test similar changes for other upload methods, so I hesitated with this PR. But I figured that might be a good idea to share my changes anyway so you might have suggestions and ideas, and then I or Brad or Lorena could move forward with finalising this, or rewriting, or continuing towards #1808

### DIFF
--- a/bcbio/qc/multiqc.py
+++ b/bcbio/qc/multiqc.py
@@ -68,7 +68,7 @@ def summary(*samples):
                 if "summary" not in data:
                     data["summary"] = {}
                 data["summary"]["multiqc"] = {"base": out_file, "secondary": data_files}
-                file_list_final = _save_uploaded_file_list(samples, file_list, out_dir)  # keep this line after out.append([data])
+                file_list_final = _save_uploaded_file_list(samples, file_list, out_dir)
                 if file_list_final:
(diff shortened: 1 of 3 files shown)


chapmanb:
Vlad;
Thanks for this, it looks great. I'm happy to have a filesystem only implementation at this point and we could extend to S3/iRODs and friends over time.

One small suggestion, what do you think about making the paths in the final multiqc file relative to the project directory? You may already do this but I didn't spot it from reading the code. This way things could be a little more "relocatable" if you need to move/combine multiple samples into another directory structure.

Thanks again for doing this.

vladsaveliev:
Thanks for a suggestion! Made the paths relative to the upload root.

chapmanb:
Brilliant, thanks again for tackling this. Much appreciated.

PR #1827, opened 2017-02-22
Update macs2.py with dm3 genome size

apastore:


chapmanb:
Alessandro -- thank you for the addition.

PR #1840, opened 2017-03-01
Fix coverage QC metrics for WGS

vladsaveliev:
This PR fixes the issue of WGS runs missing the QC metrics. 

The problem is that metrics are only added if `"base"` is in the `out` dict, whereas `"base"` is added only if either `goleft indexcov` or detailed coverage is run, and both are skipped for WGS.

Plus unrelated minor fix in multiqc.


chapmanb:
Vlad -- thanks so much for catching this and for the fix.

PR #1842, opened 2017-03-05
Qualimap file upload improvements

vladsaveliev:
- A small fix in Qualimap files uploading in order to make MultiQC happy when displaying sample names in Qualimap plots.


vladsaveliev:
Few more changes:
- Added a CWL-style file dict into RNAseq Qualimap method, on the model of the DNAseq version.
- Turning back uploading Qualimap images and HTML reports (it was removed in the DNAseq version after switching to the CWL-style dicts)

chapmanb:
Vlad, thanks so much for the all the updates and fixes with the new-style specifications. Much appreciated.

Issue #1850, opened 2017-03-08
AttributeError: 'module' object has no attribute 'etree' when parsing qsignature


lbeltrame:
An almost up to date pipeline yields this when handling qsignature:

```python
[2017-03-08T10:19Z] hibari: Unexpected error
Traceback (most recent call last):
  File "/opt/bcbio/anaconda/lib/python2.7/site-packages/bcbio/distributed/ipythontasks.py", line 50, in _setup_logging
    yield config
  File "/opt/bcbio/anaconda/lib/python2.7/site-packages/bcbio/distributed/ipythontasks.py", line 283, in qsignature_summary
    return ipython.zip_args(apply(qsignature.summary, *args))
  File "/opt/bcbio/anaconda/lib/python2.7/site-packages/bcbio/qc/qsignature.py", line 110, in summary
    out_warn_file, samples[0][0])
  File "/opt/bcbio/anaconda/lib/python2.7/site-packages/bcbio/qc/qsignature.py", line 141, in _parse_qsignature_output
    et = lxml.etree.parse(in_handle)
```

chapmanb:
Luca -- thanks much for the report and apologies about the issue. I was fixing this from a report by @ohofmann last night so it should be good to go now. Let us know if you run into any other issues at all.

lbeltrame:
Funny, I updated but I don't see the change in the code... I wonder if it's the update playing tricks on me.

PR #1851, opened 2017-03-08
Fix lxml.etree call in qsignature

lbeltrame:
lxml.etree.parse isn't valid, therefore we just import lxml.etree as
etree.

Fixes issue #1850.


chapmanb:
Thank you for this. Apologies, I was pushing an equivalent fix at the same time. Thanks for being on top of this.

PR #1853, opened 2017-03-09
Seq2C custom options fix

vladsaveliev:
Thanks for being ahead of me Brad in this! 

A minor fix from me, to avoid overriding the options in case of presence of normals. This won't affect Marisa's and Sally's case as they use -E for germline, but good to have this fix anyway.


chapmanb:
Great catch -- thanks much Vlad.

PR #1857, opened 2017-03-14
Some improvements and cleanups in the VarScan support

cbrueffer:


chapmanb:
Thanks much for the cleanup and additional support for the minimum fraction in single sample outputs. We always appreciate cleaner code.

PR #1859, opened 2017-03-14
More VarScan improvements

cbrueffer:
- Apply options specified in the VarScan resources config section
- Centrally apply some options common to paired/single sample calling
- Add --strand-filter in the single sample case to match paired calling
- general cleanup.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,8 @@
 ## 1.0.3 (in progress)
 
+- Several updates to the VarScan support: honor options specified in the
+  resource config section; honor min_allele_frac option and set --strand-filter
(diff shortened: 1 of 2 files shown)


cbrueffer:
Both varscan.py and vardict.py implement the same _safe_to_float() function. What would be a good place to add this centrally? bcbio/utils.py sounds like the obvious place but it seems to be very path/file handling specific.

chapmanb:
Christian;
Thanks for these additional improvements to VarScan and for the heads up on the duplicated safe_to_float code. I moved this over into utils where I agree it fits best. Thanks again.

PR #1862, opened 2017-03-15
Improve resources example; fix a typo.

cbrueffer:
- Quote options in the example to be safe.  Not an issue here, but
  specifying numeric arguments unquoted can lead to type mismatch
  problems.
- Add one more option to the example to make clear how to specify
  multiple options.

Specifically the vardict.py crashes with an unquoted options line like this: [-r, 2, -q, 20]


chapmanb:
Christian;
Thanks for this. I also pushed a fix to VarDict to avoid needing to quote numerical options in the specification. Much appreciated.

PR #1870, opened 2017-03-20
Handle non-string resource options similar to VarDict.

cbrueffer:


chapmanb:
Christian -- thanks much for catching all these potential issues.

PR #1875, opened 2017-03-24
Allow fractional memory specifications (e.g., 7.6G) in convert_to_bytes()

cbrueffer:
Fractional memory specifications work in all other cases I've seen. With this they also work in e.g. the BAM to FASTQ code path.


chapmanb:
Christian -- thanks much for generalizing this.

PR #1889, opened 2017-04-06
Honor additional options for hisat2 and VarDict in the RNA-seq workflow

cbrueffer:
Also clarify the docs a little.


roryk:
Thank you for these improvements, Christian, this looks great!

PR #1902, opened 2017-04-17
Unconditionally use the splice-site file.

cbrueffer:
While HISAT2 can in principle be used to align DNA reads, it is an
aligner for RNA-seq data; thus it makes sense that additional
splice-site information is always used.

This makes it easier to use the HISAT2 in the variant2 workflow
and custom workflows.

### DIFF
--- a/bcbio/ngsalign/hisat2.py
+++ b/bcbio/ngsalign/hisat2.py
@@ -28,13 +28,13 @@ def align(fastq_file, pair_file, ref_file, names, align_dir, data):
     # if assembling transcripts, set flags that cufflinks/stringtie can use
     if dd.get_transcript_assembler(data):
         cmd += "--dta-cufflinks "
-    if dd.get_analysis(data).lower() == "rna-seq":
-        gtf_file = dd.get_gtf_file(data)
-        splicesites = os.path.join(os.path.dirname(gtf_file),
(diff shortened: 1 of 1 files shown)


schelhorn:
Hi @cbrueffer, are you using HISAT2 for RNA-Seq variant calling? If so, which caller and reference assembly are you using, and how are your experiences concerning the quality of the calls?

cbrueffer:
Indeed, HISAT2 + VarDict for RNA-seq calling. Judging from a few test samples we get good calls with filtering applied. I'm calling variants in a large number of breast cancer samples at the moment, can hopefully say more in a couple of weeks. Note that there appears to be a problem with HISAT2+VarDict for calling InDels, AstraZeneca-NGS/VarDictJava#66 (possibly fixed in VarDict-Java 1.5.0).

schelhorn:
Excellent, thanks. Please report any insights, I would like the RNA-Seq variant calling pipeline to mature and your setup is my best bet for that as well.

cbrueffer:
Ping @roryk in case you missed this :-)

cbrueffer:
Turns out this is a bad change; HISAT2 is perfectly fine for DNA alignment. This is doubly so with the extended genotyping functionality in HISAT2 2.1.0.

PR #1913, opened 2017-04-24
fix for esp maf calculation

matthdsm:
Hi Brad,

We noticed that the function to calculate the ESP maf outputs some faulty results. I pushed a little fix for the function, and modified the name of the function to be more explanatory.

Let me know if you have any comments.

Cheers
M

### DIFF
--- a/config/vcfanno/hg38-gemini.conf
+++ b/config/vcfanno/hg38-gemini.conf
@@ -1,16 +1,16 @@
 [[annotation]]
 file="variation/exac.vcf.gz"
-fields = ["AC_Adj", "AN_Adj", "AC_AFR", "AN_AFR", "AC_AMR", "AN_AMR", "AC_EAS", "AN_EAS", "AC_FIN", "AN_FIN", "AC_NFE", "AN_NFE", "AC_OTH", "AN_OTH", "AC_SAS", "AN_SAS", 
+fields = ["AC_Adj", "AN_Adj", "AC_AFR", "AN_AFR", "AC_AMR", "AN_AMR", "AC_EAS", "AN_EAS", "AC_FIN", "AN_FIN", "AC_NFE", "AN_NFE", "AC_OTH", "AN_OTH", "AC_SAS", "AN_SAS",
           "AC_Het", "AC_Hom"]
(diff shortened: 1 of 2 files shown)


matthdsm:
ping @tomiles
ping @brentp, since you have the same config in the vcfanno repo

brentp:
unless I misunderstand,
for gemini, we want to know if the alternate is rare. if it were in 99.9% of individuals, this would give it a maf of 0.001 which would make it appear rare when it is common.

tomiles:
This request handles 2 issues:

- the definition of maf = the freq of the second most common allele, irrespective if that is the reference. This makes the results of the new function equivalent to how ESP itself calculates the freq in the MAF field of its vcf. (alternatively we could just parse the MAF field instead of calculating it ourselves)

- the ratio formula was not calculated correctly because the order of ref and alt counts was the reverse of what the VCF header specified it is.

chapmanb:
Matthias -- I'm also confused at the intent of this, could you explain in more detail? In general we want a stable representation of alt frequency in a population so I'm not sure why we would swap ref/alts for this calculation.

brentp:
the ESP header I'm looking at says:
```
##INFO=<ID=EA_AC,Number=R,Type=String,Description="European American Allele Count in the order of RefAllele,AltAlleles. For INDELs, A1, A2, or An refers to the N-th alternate allele while R refers to the reference allele.">
```
so the first allele is always the ref. To get alternate alelle count, we would then do `v[2] / (v[1] + v[2])`, yes? 

if you want minor allele frequency, your change is correct, but in gemini, we want alternate allele frequency.

matthdsm:
That's weird....
The ESP file I have here contains this header:
```
##INFO=<ID=EA_AC,Number=.,Type=String,Description="European American Allele Count in the order of AltAlleles,RefAllele. For INDELs, A1, A2, or An refers to the N-th alternate allele while R refers to the reference allele.">
```

I'm using the ESP vcf file that was installed by the bcbio installer. It seems like there's some kind of mixup between files then..

M

tomiles:
The vcf from the http://evs.gs.washington.edu/EVS/ download section has the header with the order of AltAlleles,RefAllele
=> http://evs.gs.washington.edu/evs_bulk_data/ESP6500SI-V2-SSA137.GRCh38-liftover.snps_indels.vcf.tar.gz

brentp:
oh, I see. we're still using the old 37 version. We should verify with the UW folks that what's in the header is correct as that is suprising. 
@jxchong, do you know about this?

matthdsm:
I just opened a new PR, where I've switched the values in the calculation. Everything should be correct now.

Cheers
M

PR #1915, opened 2017-04-24
updated ESP AF calculation

matthdsm:
related to #1913

### DIFF
--- a/config/vcfanno/hg38-gemini.lua
+++ b/config/vcfanno/hg38-gemini.lua
@@ -63,9 +63,9 @@ function div2(a, b)
 end
 
 function ratio(vals)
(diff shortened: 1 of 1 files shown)


tomiles:
I checked a few vcf records and the new order AltAlleles,RefAllele in the header for EA_AC,AA_AC and TAC agrees with the order used in the actual variants rows.
So this PR does fix the proper calculation of the freq/ratio for the new ESP6500SI-V2-SSA137.GRCh38-liftover version.

chapmanb:
Thanks for the fix, it's frustrating to have different order for different builds but glad we now that the right thing in place. Thanks again for spotting this.

matthdsm:
No problem, happy this got resolved!
Thanks
M

PR #1922, opened 2017-05-05
update config docs PE auto config

stephenturner:
add documentation on problem with sequentially numbered samples


lpantano:
Thanks!

PR #1926, opened 2017-05-09
vep update

matthdsm:
- updated genome configs for vep v88
- updated effects.py for vep v88
- added vcfanno config for dbNSFP and dbscSNV


matthdsm:
related to chapmanb/cloudbiolinux#250

matthdsm:
Hi Brad,
quick question. Since the PR for cloudbiolinux was merged, new installations will only get the latest versions of the reference data files. For clinvar there's no issue, but won't this cause conflicts for 'fresh' installs of the genome resources here aren't updated too? This is the case for the vep cache and the dbsnp file, since the name changed too.

Cheers 
M

chapmanb:
Matthas;
Agreed, thanks for the push. Apologies, I'd done some testing but hadn't had a chance to finish. Thank you again for all this work getting it in place. We'll aim for a soonish release after folks have a chance to test so we can get data and code all synchronized in a stable release. Thank you again.

matthdsm:
Hi Brad,

My pleasure. If you're interested, I'm willing to commit to keeping the datasets we use up to date. Is it possible for me to get some kind of notification if you're about to release a new version? That way I can check if everything's still up to date and prepare the PR's needed to include the latest data in the new release.

Cheers
M

chapmanb:
Matthias -- thanks so much. I'll try to provide a heads up when we're reaching time for a new release. We're aiming to keep bcbio stable in development and just break off new releases over time. In general we try to get a release out in a month/two-month time frame so practically anytime around then is reasonable to start putting this together. I'd suggest only doing updates when you think it provides some value for interpretation, since there is a cost to needing to re-download and validate everything for users. Thank you again.

matthdsm:
No problem,
judging by past releases, I'll start looking into new datasets every two months, at the beginning of the month, and put together a PR.

Cheers
M

PR #1932, opened 2017-05-16
Handle empty resource sections instead of crashing.

cbrueffer:
Previously having the following snippet in the config would lead to a crash:

```
resources:
  default:
    # no specifications here
```


chapmanb:
Christian, thanks much for catching this and for the fix.

PR #1937, opened 2017-05-17
Fix cnvkit issue on directory names with dots

vladsaveliev:
This part of CNVkit code removes dots in full file names leading to problems when a run directory name contains dots. Changing it to only remove the last dot in common prefix. Hope this is what was originally intended in this code!


chapmanb:
Vlad -- thanks for catching this and for the fix. That's exactly the intent of the code, sorry for missing this case.

PR #1950, opened 2017-05-24
Fix Seq2C for paired samples

vladsaveliev:
When running on tumor/normal pairs, Seq2C's cohort calling function returns a cut sample list containing only tumors, leading to skipping normals in downstream steps, ending with no normals in`final`. This PR fixes that.


chapmanb:
Vlad -- thanks for catching this and for the quick fix.

PR #1953, opened 2017-05-25
Seq2C coverage file is not batch-level

vladsaveliev:
Hi Brad, another minor fix for Seq2C. The `*-coverage` files it reports are not batch-level, but sample-level, so they should probably have the sample names in their files names rather than batch names. Basically that's not a big deal to just keep it as batch (it breaks postproc but easily fixable), but I figure it would be nice to correct it in bcbio.


chapmanb:
Vlad -- thanks so much for being on top of this. Much appreciated.

PR #1970, opened 2017-06-09
bumped vep cache versions

matthdsm:


chapmanb:
Matthias;
Thanks much for this. I also bumped all of the resource file versions in the YAML so that bcbio upgrades will know there is a new version of these files and download them. Thanks again.

matthdsm:
Hi Brad, 

Thanks, I'll keep it in mind to bump the resource file versions too in the future.
Cheers
M

PR #1975, opened 2017-06-12
VEP tweaks

matthdsm:
Added support for new SpliceRegion plugin
Plugin file already added to conda recipe.

Cheers
M


matthdsm:
I also added some flags to to stop vep from complaining there's an newer subversion available.
The command will now also overwrite older converted files to prevent issues with the tabix index.
After conversion, all converted files will now also be removed to prevent VEP from using them, causing memory issues down the line.

Cheers
M

matthdsm:
I realized the vep_install command doesn't support the flags mentioned above, so I restored the previous convert_cache command, as it seems to perform better for some reason and supports the newly added flags.

M

chapmanb:
Matthias -- thanks so much for this and sorry for the delay in merging, I just realized I forgot to pull this in. I appreciate you being on top of all of this.

PR #1995, opened 2017-06-30
simplified dbnsfp vcfanno conf

matthdsm:
simplified `ops` for dbsnfp vcfanno conf, killing of a bunch of errors in the process.
Changes all ops to `first` since dbNSFP only contains one line per position.

M


chapmanb:
Matthias -- thanks much for cleaning this up.

PR #2017, opened 2017-07-21
Clean up unused imports.

cbrueffer:


cbrueffer:
The Travis check failure seems unrelated to the change.

chapmanb:
Thanks much for the cleanup. Tests are having some issues right now due to running out of space on the TravisCI instance, so that is unrelated. Thanks again.

PR #2034, opened 2017-08-07
Add singlecell-barcodes installation.

roryk:
Some of the common barcode schemes have a large amount of possible cellular barcodes, so I moved these out of the bcbio-nextgen repo and into its own standalone one to reduce clutter. @chapmanb my spidey sense is going off that this wasn't how you wanted stuff added, so pinging you to let me know if there is another way of doing it that works better. ggd-recipes didn't seem like a great solution since it isn't genome specific.


roryk:
Moved this over to a bioconda recipe instead. Thanks @chapmanb for the good suggestion.

PR #2036, opened 2017-08-09
fixed faulty column numbers in dbnsfp vcfanno configs

matthdsm:
fixed issue with column numbering in vcfanno config, where the wrong values would be in the wrong column.

Cheers
M


chapmanb:
Matthias -- thanks much for the fix.

PR #2042, opened 2017-08-20
Aggregate MultiQC configuration in one place 

vladsaveliev:
Moving `table_columns_visible` and `module_order` settings into MultiQC_bcbio: https://github.com/MultiQC/MultiQC_bcbio/pull/14. Except for those that are set dynamically (e.g. when `qualimap` is on).

Also, `"samtools": {"error_rate": False}}` doesn't work - should be `"Samtools Stats": {"error_rate": False}}`.


chapmanb:
Thanks Vlad for centralizing this. I'm working on updating the multiqc-bcbio bioconda package with your fixes there and then will get this merged in.

chapmanb:
Thanks again for the code cleanup here.

PR #2043, opened 2017-08-20
Add Preseq tool for complexity prediction.

vladsaveliev:
Runs [`preseq lc_extrap`](http://smithlabresearch.org/software/preseq/) that estimates the yield for future experiments by showing how many additional unique reads are sequenced for increasing total read count. 

Activated by adding `preseq: true` to the `algorithm` section. I wasn't sure if it should be done using `tools_on` or with its own parameter. However, I want to add another preseq command `c_curve` for yield estimation of lower input experiments, and it would make sense to support configuration like `preseq: lc_extrap` and `preseq: c_curve`.

Reasonable prediction parameters are calculated (estimation step and limit) based on the total read count, but also configurable in `resources` section (I described it in the docs).

To aid visibility through the MultiQC plot, bcbio adds some metadata into MultiQC config and writes a file with actual counts. It makes MultiQC convert the molecule numbers into depths of coverage, and display dots with actual coverages (see examples in this comment https://github.com/ewels/MultiQC/pull/461#issuecomment-301625384).

I'm not sure if I did everything right with cwl compatibility. Hope it works well, but I'll try to install bcbio-vm eventually to test it myself.

### DIFF
--- a/bcbio/cwl/defs.py
+++ b/bcbio/cwl/defs.py
@@ -327,7 +327,7 @@ def variant(samples):
                             cwlout(["summary", "metrics"], "string"),
                             cwlout("inherit")])],
             "bcbio-vc", ["bcftools", "bedtools", "fastqc", "goleft", "picard", "pythonpy",
-                         "qsignature", "qualimap", "sambamba", "samtools"]),
+                         "qsignature", "qualimap", "sambamba", "samtools", "preseq"]),
           s("multiqc_summary", "multi-combined",
             [["qcout_rec"]],
             [cwlout(["summary", "multiqc"], ["File", "null"])],
--- a/bcbio/pipeline/main.py
+++ b/bcbio/pipeline/main.py
@@ -145,7 +145,7 @@ def variant2pipeline(config, run_info_yaml, parallel, dirs, samples):
     with prun.start(_wres(parallel, ["gatk", "gatk-vqsr", "snpeff", "bcbio_variation",
                                      "gemini", "samtools", "fastqc", "sambamba",
                                      "bcbio-variation-recall", "qsignature",
-                                     "svcaller"]),
+                                     "svcaller", "preseq"]),
                     samples, config, dirs, "multicore2",
                     multiplier=structural.parallel_multiplier(samples)) as run_parallel:
         with profile.report("joint squaring off/backfilling", dirs):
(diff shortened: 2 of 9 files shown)


vladsaveliev:
By the way, the usage scenario I have in mind is to estimate quality of experiments based on a fraction of input, e.g. randomly take 5% of reads (I guess using a function like https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/bam/fastq.py#L190) and run bcbio with preseq, qualimap, and fastqc on, and infer future full experiment's stats based on the downsampled results (e.g. preseq will help to predict the duplication rate and coverage depth). Brad, I'm wondering if you already did something like that in your experience and if bcbio can be configured to downsample reads to a certain percent?

chapmanb:
Vlad -- awesome, thanks so much for all these PRs. I'll get them pulled in tomorrow along with the MultiQC updates.

We do have downsampling along the lines of what you describe for FastQC. There we downsample to a target number of reads and bcbio calculates the percentage and runs the downsampling:

https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/bam/__init__.py#L171

For CWL we'll need to eventually have a single point for doing this and also separate out QC into those that only require BAMs and those that need VCFs and other outputs so the pre-check style runs can happen earlier in the process. But that's a longer term goal.

Thanks again for all this work.

vladsaveliev:
Brad, that `downsample` function seem to be running from BAM files only. So the alignment needs to be done first, however my goal is to avoid full alignment for quick downsampled QC. That's why I'm looking at the function that downsamples fastq https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/bam/fastq.py#L190. However it doesn't seem to get called anywhere in the code. I'm using a similar code in TargQC for quick downsampled QC, but now I'm thinking about using bcbio for that.

chapmanb:
Vlad -- sorry my fault for pointing in the wrong direction. My suggestion would be to use `seqtk sample` for the downsampling. It can sample by either number or fraction of reads, although we might want to stick with a defined number that works fairly quickly so we can handle cases with lower numbers of reads as well.

If we're not using the fastq downsampling in bcbio it would be worth removing that for clarity and swapping over to use the `seqtk` approach. Thanks again for looking at this.

vladsaveliev:
Thanks Brad, sounds reasonable! I'll be working on that and will do a PR if I end up updating the downsampling function.

chapmanb:
Brilliant, thanks Vlad. I'll get this in to get things started and we can update from here. Much appreciated.

Issue #2045, opened 2017-08-24
Change MultiQC plot limit


schelhorn:
In MultiQC, there exists an option `num_datasets_plot_limit` to control the number of samples that will result in a static rather than a dynamic plot. Is there a way to control this cutoff in bcbio, perhaps as a MultiQC `resource.yaml` option?

chapmanb:
Sven-Eric thanks for the discussion. I'll have to ping @vladsaveliev to see what he thinks the best approach is to customize MultiQC parameters. Vlad, I know you were just re-arranging configuration -- did you have any plans about how to let folks tweak those? Alternatively, maybe we could just have a better default for `num_datasets_plot_limit` in bcbio that matches what folks want. Sven-Eric, what did you want to adjust it to?

vladsaveliev:
Hi Sven-Eric, Brad,

There are multiple options to configure MultiQC: http://multiqc.info/docs/#configuring-multiqc and two of them look suitable for bcbio:
- Put `~/.multiqc_config.yaml` into your home directory,
- Point `MULTIQC_CONFIG_PATH` environment variable to the location of your yaml.

For example, if you create a file `~/.multiqc_config.yaml` with the following contents:
```yaml
num_datasets_plot_limit: 100
```
MultiQC that is launched in bcbio will pick it up correctly and use your parameter instead of defaults.

By the way, in order to set a sample limit to render static plots you need another option `plots_flat_numseries` (see http://multiqc.info/docs/#flat--interactive-plots). `num_datasets_plot_limit` does a different thing and supposed to disable automatic loading of plots: http://multiqc.info/docs/#disabling-on-load-plotting

vladsaveliev:
However we can actually expose the [`--cl_config` option](http://multiqc.info/docs/#configuring-multiqc) in the bcbio yaml resources section. It will allow to set any kind of MulitQC parameters directly in bcbio yaml, and it's always good to centralize configuration.

lpantano:
That is a good idea, maybe with `resources`: `multiqc`: `options`: [—cl ….], we’ll just need to check if there are options and put it when we execute multiqc.

schelhorn:
The proposal to put it into the resource section seems to be most elegant, in my opinion. That's also how it works for other executables used by bcbio such as `vardict`, for instance.

vladsaveliev:
Agree. Added this feature, MultiQC can be configured via the resources section, e.g.:
```
resources:
  multiqc:
[...]

PR #2048, opened 2017-08-25
Multiqc config

vladsaveliev:
MultiQC:
1. upload multiqc_config.yaml into final to allow reruns.
2. support custom options in MultiQC resources (#2045), e.g.:
```
resources:
  multiqc:
    options: ["--cl_config", "'qualimap_config: { general_stats_coverage: [20,40,200] }'"]
```
Perhaps it makes sense to document it in `configuration.rst`, but not sure of the best way.


chapmanb:
Thanks so much for doing this Vlad. I pushed a small doc update to include this in the QC section of the configuration. Much appreciated.

PR #2050, opened 2017-08-28
updated vep cache version

matthdsm:
Updated required vep cache version


chapmanb:
Matthias -- thanks again for keeping these up to date. When you update the configuration file you also need to manually increment the versions in the YAML files so bcbio knows to download the new versions on existing installs. I did that in a separate commit following on this so we should be good to go now. Thanks again.

PR #2057, opened 2017-09-02
Kraken: do not require BAMs or aligners to run

vladsaveliev:
Not the most elegant solution, but since we already have a [special check](https://github.com/chapmanb/bcbio-nextgen/blob/072f90f9a803bfde1449ab80c3665fae4515c617/bcbio/ngsalign/alignprep.py#L228) for kraken to keep the ogirinal input files, I hope it doesn't look too bad to treat kraken specially in QC as well.
cc @lpantano


lpantano:
It looks good to me.Thanks for this!

chapmanb:
Vlad -- thanks much for this work. Lorena -- thanks for the review. Let us know if you run into anything else.

PR #2060, opened 2017-09-06
VEP: replaced ExAC with GnomAD

matthdsm:
The `--af_exac` flag had been deprecated in favor of the `--af_gnomad` flags, which includes annotations of all GnomAD exome frequencies. Using the `--af_exac` flag breaks VEP, even though it's still referenced in the docs.
see ensembl/ensembl-vep#76

### DIFF
--- a/bcbio/variation/effects.py
+++ b/bcbio/variation/effects.py
@@ -165,7 +165,7 @@ def run_vep(in_file, data):
                         "--offline", "--dir", vep_dir,
                        "--symbol", "--numbers", "--biotype", "--total_length", "--canonical",
                        "--gene_phenotype", "--ccds", "--uniprot", "--domains", "--regulatory",
-                       "--protein", "--tsl", "--appris", "--af", "--max_af", "--af_1kg", "--af_esp", "--af_exac",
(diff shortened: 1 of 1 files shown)


chapmanb:
Matthias -- thanks much for catching this and for the fix. How are things running generally with VEP 90? I'm hoping to have a new release soon and wanted to make sure it was in a good state for you before rolling it out.

matthdsm:
Hi Brad,

No problem. We're currently running a big analysis with v90 and expect to have results later this week.
I'll keep you up to date on any issues we're having. If you don't hear from me you can assume everything works fine.

Cheers
M

matthdsm:
Hi Brad,

VEP90 seems to work fine. It seems however, the update causes some issues downstream with vcf2db. It seems vcf2db is having a hard time parsing the CSQ tag. I've opened an issue about this [here](quinlan-lab/vcf2db#32)

Cheers
M

chapmanb:
Matthias -- thanks much, I'll follow the issue and bump vcf2db after Brent or Aaron has a chance to look.

PR #2063, opened 2017-09-06
added vcfanno config for GnomAD genome vcf

matthdsm:
Hi Brad,

I've added a first draft of a vcfanno config for GnomAD. Please feel free to merge and edit as you see fit. As it is now, I just copied all allele frequency related tags from the vcf header, without any further processing.

Cheers
M


matthdsm:
The latest push restricts the annotation fields to AF only, since the rest is mostly additional data which can be looked up in the gnomad browser. I also changed the field names to reflect the fact the annotations come from the genome dataset.

M

chapmanb:
Matthias;
Thanks much for these. Could we name the INFO field annotations just `gnomAD_af_*` instead of `gnomAD_genome_af_*`? I know you're looking to distinguish them but this seems a little long for a key name. Would that work okay with current GEMINI gnomAD support?

matthdsm:
Hi Brad,

We're going to need to differentiate between exome/genome annotations, since VEP uses the same naming scheme. If I leave out the `genome` part, we'd have identical field names. Perhaps I can shorten it to `gnm` instead? That way, the names wont be much longer than the ExAC ones.

As for Gemini, I've no idea if this will be compatible. We don't use is anymore since migrating to hg38. I added to config just for completeness.

Cheers
M

chapmanb:
Matthias;
Thanks much for the discussion. Does it matter if we use the same key names as VEP and take precedence over them? From what I understand VEP only offers gnomAD annotations in exome regions to save space, so these should be identical in those cases, but be more complete over the whole genome. Is there a value in having both side by side?

matthdsm:
Good question, 
I hadn't thought about it like that.
I think you're right, but I've no idea how vcf2db will handle this, since vcfanno add separate INFO tags and VEP pushes everything in a CSQ tag. The data from the CSQ tag also gets parsed further for the variant impacts table. I fear this will cause some conflicts in the script, or worse, mixups in the data.

chapmanb:
From my understanding of vcf2db, I believe the INFO fields should take precedence but the data should be the same either way so as long as it includes them we'll be fine. Could we take the same name approach as a starting point and then investigate vcf2db issues if we run into them? I'd rather keep the vcfanno output smaller and easier to understand.

matthdsm:
Hi Brad,

I've adjusted the field names and made them to match with the ones added by Ensembl. As you said, we can take it from here and see if any issues arise down the road.

Cheers
M

chapmanb:
Matthias -- thanks for this. Unless I'm misunderstanding the config, it would be worth adjusting this to be separate key values rather than one long key and values. Matching the names is great but having them separate would make it easier for users to parse downstream. Thanks again.

matthdsm:
whoops, my bad! Copy paste error! I'll fix this.

M

chapmanb:
Brilliant, thanks again for this.

PR #2072, opened 2017-09-13
removed clinvar from dbNSFP to avoid confusion

matthdsm:
Hi Brad,

I've removed the clinvar related fields from dbNSFP. My reasoning behind this is the following:
- we already have clinvar annotation through VEP/gemini/vcfanno-gemini
- the existing clinvar annotations are more recent than those in dbNSFP
- having a multitude of clinvar related fields with similar names causes confusion about which is which.

Hope you agree, if not, feel free to close this PR.

Cheers
M


chapmanb:
Matthias -- thanks for this. This makes good sense and I agree with having the clinvar fields directly from the VCFs.

PR #2073, opened 2017-09-13
QC: merge padded bed file

vladsaveliev:
Minor fix for the reads-on-padded-target metric


chapmanb:
Vlad -- thanks much for the fix.

PR #2096, opened 2017-10-11
Clean up MultiQC for germline variants calling

vladsaveliev:
- When germline calling requested for "normal" samples, write germline VCF into the "normal" results folder rather than creating a separate `<normal-sample>-germline` folder with duplicated BAM and stats.
- Run Bcftools twice, once for somatic and once for germline calls; both for the paired case (when germline calling was performed on normals), and for the single tumors (when somatic prioritization was done with the following extraction of germline variants). 
- Report both of the Bcftools runs in MultiQC (needs MultiQC PR https://github.com/ewels/MultiQC/pull/608 to be accepted).
- Remove the germline call stats from the General Stats MultiQC table, and rather report them in a separate table right above the germline Bcftools section (also needs the MultiQC PR):
screen shot 2017-10-11 at 16 50 32

Addresses the clean up part at https://github.com/chapmanb/bcbio-nextgen/issues/2081

### DIFF
--- a/bcbio/pipeline/qcsummary.py
+++ b/bcbio/pipeline/qcsummary.py
@@ -139,6 +139,8 @@ def _run_qc_tools(bam_file, data):
     for program_name in dd.get_algorithm_qc(data):
         if not bam_file and program_name != "kraken":  # kraken doesn't need bam
             continue
+        if dd.get_phenotype(data) == "germline" and program_name != "variants":
+            continue
         qc_fn = tools[program_name]
         cur_qc_dir = os.path.join(qc_dir, program_name)
         out = qc_fn(bam_file, data, cur_qc_dir)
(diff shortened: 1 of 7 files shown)


schelhorn:
Great!

chapmanb:
Vlad, awesome thanks for this. I'm not totally sure I understand what happens now with the germline naming but left a note and will read again. I was using `-germline` to avoid collisions with identical callers for both germline and somatic. Does this still handle that case?

vladsaveliev:
Brad, somatic calling is done only for tumors, and germline calling is done only for normals, so the collision is resolved just based on the sample names. The "germline" settings does not affect unpaired samples. Finally, for "tumor" unpaired samples, the `-germline` suffix is used when germline calls are extracted.

chapmanb:
Sounds great, thanks Vlad. The tests are becoming unresponsive now for an unknown reason, so that failure is unrelated and we'll get all this in and start testing. Thank you again.

lbeltrame:
Does this pick up the data from the `germline: [<variantcaller>]` results?

PR #2109, opened 2017-10-16
added idxstats to samtools qc step

matthdsm:
Hi Brad,

I took the liberty of adding a little QC step to the samtools routine. Using Idxstats in multiqc enables us to quickly check for the gender of samples based of the XY read count.

This is just a basic implementation, and I'm not sure of more steps should be added. I was hoping you could provide further plumbing if necessary.

Thanks a lot!
M


chapmanb:
Matthias -- brilliant, thank you for adding this in. It's always great to have more thorough MultiQC output and ways to debug gender issues. Much appreciated.

PR #2111, opened 2017-10-17
Fix uploading germline VCFs for normals

vladsaveliev:
Thanks @ohofmann for spotting the issue.

Changing `<normal>-germline` to `<normal>` lead to an issue during one of the SV callings steps where the samples are grouped by name. This part of the code wasn't run in by my test cases so it wasn't cought earlier unfortunately.

Fixing that now.


chapmanb:
Vlad -- briliant, thanks so much for diagnosing and the quick fix.

PR #2114, opened 2017-10-18
Put back the workaround from 8fe8b41 for gatk4 and applied to gatk3.8

biocyberman:
This workaround first appeared on 8fe8b41 but then got removed without tracking
on the next commit 7ec8a71. GATK 3.8 is also affected by the same problem and
the workaround works fine for it.

### DIFF
--- a/bcbio/variation/recalibrate.py
+++ b/bcbio/variation/recalibrate.py
@@ -127,10 +127,10 @@ def _gatk_apply_bqsr(data):
             if gatk_type == "gatk4":
                 params = ["-T", "ApplyBQSRSpark", "--sparkMaster", "local[%s]" % cores,
                           "--input", in_file, "--output", tx_out_file, "--bqsr_recal_file", data["prep_recal"],
-                          "--conf", "spark.local.dir=%s" % os.path.dirname(tx_out_file)]
+                          "--conf", "spark.local.dir=%s" % os.path.dirname(tx_out_file), "--use_jdk_deflater"]
             else:
                 params = ["-T", "PrintReads", "-R", dd.get_ref_file(data), "-I", in_file,
-                          "-BQSR", data["prep_recal"], "-o", tx_out_file]
+                          "-BQSR", data["prep_recal"], "-o", tx_out_file, "--use_jdk_deflater"]
             memscale = {"magnitude": 0.9 * cores, "direction": "increase"} if cores > 1 else None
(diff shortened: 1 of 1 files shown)


chapmanb:
Thanks for the review on this. We removed the workaround in a later commit with documentation here:

https://github.com/chapmanb/bcbio-nextgen/commit/ca47837798f0f2d86de44441a898a4cc03925f42#diff-24cb97d2a3cdff87295b4268f26bb914

after reporting and discussing with the GATK team (https://github.com/broadinstitute/gatk/issues/3605) where the issue was due to simulatenous use of SerialGC in the java options. By removing that we were able to resolve the underlying issue without needing to avoid the GKL deflater, but it's only applied when running GATK4 with multiple cores which is where we reproducibly found the issue. Could you provide more details about the failure you're seeing and happy to debug more and see if we need to swap this back in or can use a different approach? Thanks much.

biocyberman:
I am running exome analysis for Illumina Nextseq PE data. Yaml config file is provided in the attached tar.gz file.

[project11_debug.tar.gz](https://github.com/chapmanb/bcbio-nextgen/files/1398325/project11_debug.tar.gz)

bcbio_nextgen failed at this step:

`unset JAVA_HOME && export PATH=/bcbio/anaconda/bin:$PATH && /bcbio/anaconda/bin/gatk -Xms750m -Xmx104833m -XX:+UseSerialGC -Djava.io.tmpdir=/data/project11/work/tmp/tmpxi55a_ -T PrintReads -R /bcbio/genomes/Hsapiens/hg19/seq/hg19.fa -I /data/project11/work/align/Daughter/Daughter-sort.bam -BQSR /data/project11/work/align/Daughter/Daughter-sort-recal.grp -o /data/project11/work/tmp/tmpxi55a_/Daughter-sort-recal.bam -U LENIENT_VCF_PROCESSING --read_filter BadCigar --read_filter NotPrimaryAlignment`

Java core dump was produced after I set `ulimit -c unlimited`. It is very similar to the one you report for GATK4. You can find the core dump file in the attached compressed log files. 

By adding the `--use_jdk_deflater` I could get pass this point. The whole run is still running.

chapmanb:
Thanks much for all the details, this helped a lot to understand the context. I pushed a fix to hopefully avoid this by skipping the incompatible SerialGC option in this case for both GATK3 and GATK4. I also pushed a similar fix to VQSR where I've seen similar problems reported (thanks @ohofmann). Hopefully this fixes the issue for you and allows it to proceed cleanly.

PR #2116, opened 2017-10-19
Fix SV ensemble batching with germline sample (#2111 part 2)

vladsaveliev:
One more fix for that. Address the scenario of ensemble SV calling + tumors sharing one normal + germline calling for normal.


chapmanb:
Thanks again for identifying and fixing all these issues.

PR #2119, opened 2017-10-23
Fixed missing import of bio.structrural.shared

biocyberman:
Line 131 in cnvkit.py calls shared.find_case_control() but `shared` module was
not imported.


chapmanb:
Vang -- thank you for catching this. Much appreciated.

biocyberman:
Hi Brad
`cnvkit` failed, and I could see why, therefore a tiny fix could help me going :)

PR #2121, opened 2017-10-23
Ensured VCF file input for metasv.py is not None.

biocyberman:
`metasv.py` failed. After some debugging, I could see that both `call["vcf_file"]` and `call["vrn_file"]` were `None`. Therefore this line caused an exception:

`cmd += ["--%s_vcf" % call["variantcaller"], call.get("vcf_file", call["vrn_file"])]`

### DIFF
--- a/bcbio/structural/metasv.py
+++ b/bcbio/structural/metasv.py
@@ -26,9 +26,10 @@ def run(items):
                         "--bam", dd.get_align_bam(data), "--outdir", work_dir]
     methods = []
     for call in data.get("sv", []):
-        if call["variantcaller"] in SUPPORTED and call["variantcaller"] not in methods:
+        vcf_file = call.get("vcf_file", call.get("vrn_file", None))
+        if call["variantcaller"] in SUPPORTED and call["variantcaller"] not in methods and vcf_file is not None:
             methods.append(call["variantcaller"])
-            cmd += ["--%s_vcf" % call["variantcaller"], call.get("vcf_file", call["vrn_file"])]
+            cmd += ["--%s_vcf" % call["variantcaller"], vcf_file]
     if len(methods) >= MIN_CALLERS:
         if not utils.file_exists(out_file):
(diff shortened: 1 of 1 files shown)


chapmanb:
Thanks much for the fix. As a heads up, metasv integration hasn't gotten much work recently. Definitely happy to have you revive it and fix any issues if it's useful in your work. We typically use an approach of combining callers and prioritizing in regions of interest with `svprioritize` (https://bcbio-nextgen.readthedocs.io/en/latest/contents/configuration.html#structural-variant-calling) which avoids the overhead of recalling and recalulcating breakpoints. Hope this helps.

biocyberman:
Thanks Brad. I was also read the paragraph about `svprioritize` when I was making configuration file. However, I did not understand what input file I should use. We have exome data and want to run SV detection as an additional step to variant calling. All of these items do not seem to fit to our case:



> -         cancer/civic (hg19, GRCh37, hg38) – Known cancer associated genes from CIViC.
> -         cancer/az300 (hg19, GRCh37, hg38) – 300 cancer associated genes contributed by `AstraZeneca oncology
> -         cancer/az-cancer-panel (hg19, GRCh37, hg38) – A text file of genes in the AstraZeneca cancer panel. This is only usable for svprioritize which can take a list of gene names instead of a BED file. <https://www.astrazeneca.com/our-focus-areas/oncology.html>`_.
> -         actionable/ACMG56 – Medically actionable genes from the The American College of Medical Genetics and Genomics

chapmanb:
Vang -- if you want SVs in all exons than you're right that `svprioritize` is not especially useful and the best approach is to look at the individual callers. Manta and CNVkit are good first passes to identify events, and Lumpy and WHAM are useful if you need more specificity. `svprioritize` would be useful if you have a subset of genes of interest and want to dig into the additional specificity only in those. Most of the current pre-built inputs are cancer based but if none of these fit your needs you can use a BED file of genes of interest. Hope this helps.

PR #2123, opened 2017-10-24
Fix autopair UMI tagging

lbeltrame:
`_find_umi` returned the files in the wrong order (R1, R3, R2) so R1 and
the UMI file would get tagged with reads in R2 instead.

### DIFF
--- a/scripts/bcbio_fastq_umi_prep.py
+++ b/scripts/bcbio_fastq_umi_prep.py
@@ -128,7 +128,7 @@ def _find_umi(files):
         return exts["R1"], exts["R2"], exts["I1"]
     else:
         assert "R3" in exts, exts
-        return exts["R1"], exts["R3"], exts["R2"]
+        return exts["R1"], exts["R2"], exts["R3"]
 
 def _commonprefix(files):
     """Retrieve a common prefix for files without extra _R1 _I1 extensions.


mjafin:
Can I ask to make sure this doesn't change the way UMIs are processed for current users? We get the reads in R1 and R3, UMIs in R2 coming out of bcl2fastq

lbeltrame:
At least from the way I understood the code, `r1` and `r2` are meant to be the reads, and `r3` the UMI. However I'm not using Illumina UMIs (I preprocess them from fastq). 
If this is intended, a code comment would be appropriate (but I still think this fails the principle of least surprise).

mjafin:
We're not using Illumina UMIs (i.e. UMIs in the actual reads) but rather UMIs that are in the index reads. Illumina's bcl2fastq doesn't directly support such UMIs but allows exporting them into a separate read files. What happens is you get the reads in R1 and R3, UMIs in R2. If there are dual UMIs then the reads will be in R1 and R4, UMIs in R2 and R3 (a scenario not considered by bcbio). 

A code comment would indeed be good but I wouldn't change the behaviour.

lbeltrame:
One reason I wanted to "fix" `autopair` is that `single` runs `umis` single-core. I might look into that instead (but I'll wait for Brad's input first).

chapmanb:
Luca and Miika -- thanks much for the PR and discussion. Miika is right that this is the intended behavior based on what bcl2fastq does. Apologies for the confusion; I've documented this to be more clear. I also added the ability to tweak core usage for a single run. This only affects parallel bgzip since umis is single core, and the additional parallelization in autopair is per fastq file. Hope this helps get it working for you.

lbeltrame:
Thanks for the clarification, Brad. Do you mean that the `cores` parameter of `umis fastqtransform` has little to no impact?

```
$ umis fastqtransform --help
Usage: umis fastqtransform [OPTIONS] TRANSFORM FASTQ1 [FASTQ2] [FASTQ3]
                           [FASTQ4]

  Transform input reads to the tagcounts compatible read layout using
  regular expressions as defined in a transform file. Outputs new format to
  stdout.

Options:
  --keep_fastq_tags
  --separate_cb         Keep dual index barcodes separate.
  --demuxed_cb TEXT
  --cores INTEGER
  --fastq1out TEXT
  --fastq2out TEXT
  --min_length INTEGER  Minimum length of read to keep.
  --help                Show this message and exit.
```

chapmanb:
Luca -- sorry I'd missing there is some internal parallelization in umis fastqtransform, thanks for the heads up. @roryk, do you think turning this on is valuable to improve runtimes? We're trying to decide how to allocate cores between `umis`, `bgzip` output and per-sample parallelization.

lbeltrame:
Normally it's not a problem, however I'm having some targeted very high coverage (5000+X) samples and I wonder if I could (significantly) shave off some runtime when processing UMIs.

PR #2124, opened 2017-10-24
support for BCBIOPATH: proof of concept

gberriz:
The proposed change provides support for an optional `BCBIOPATH` environment variable.

If this variable is set and non-empty, the non-empty paths it mentions are prepended to `PATH`.  (If `BCBIOPATH` is unset, this code is essentially a no-op.)

The updating of `PATH` takes place the first time that `bcbio.setpath` is `import`'ed.  (Afterwards, `bcbio.setpath` remains as an inert placeholder.)

As the message for the PR's commit indicates, this is only a proof of concept.  I don't expect it to be incorporated into master as-is, but I hope it could be incorporated with little aditional work.

The name `BCBIOPATH` is only a suggestion.  Please feel free to change it if desired.

I have done some light, preliminary testing of the modified code, and it seems to work well, but more thorough testing is in order.

### DIFF
--- a/bcbio/setpath.py
+++ b/bcbio/setpath.py
@@ -1,0 +1,60 @@
+"""Update the PATH environment variable to reflect the value of BCBIOPATH.
+
+"""
+
+import os
+
+def _maybe_prepend(original, to_prepend):
+    """Prepend unique non-empty paths unless already present.
+
+    original and to_prepend are expected to be strings corresponding to
+    os.pathsep-separated lists of filepaths.
+
+    If to_prepend is the empty sting, original is returned.
+
+    examples:
+    # Unix
+    _maybe_prepend('/foo:/bar:/foo', '/foo:/baz:/baz') -> '/baz:/foo:/bar:/foo'
+    _maybe_prepend('/baz', '/foo::/bar:/foo')          -> '/foo:/bar:/baz'
+    _maybe_prepend('/foo:/bar:/foo', '')               -> '/foo:/bar:/foo'
+
+    """
+
+    sep = os.pathsep
+
+    def split_path_value(path_value):
+        return [] if path_value == '' else path_value.split(sep)
+
+    new_paths = split_path_value(to_prepend)
+
+    if new_paths == []:
+        return original
+
+    paths = split_path_value(original)
+
+    paths_lookup = set(paths)
+
+    # collect in prefix all unique non-empty paths in new_paths that are not
+    # already in paths
+    prefix = []
+    for path in new_paths:
+        if path not in paths_lookup and path != '':
+            prefix.append(path)
+            paths_lookup.add(path)
+
+    return sep.join(prefix + paths)
+
+
+if __name__ == '__main__':
+
+    # check examples (poor-man's doctest)
+    print _maybe_prepend('/foo:/bar:/foo', '/foo:/baz:/baz')
+    print _maybe_prepend('/baz', '/foo::/bar:/foo')
+    print _maybe_prepend('/foo:/bar:/foo', '')
+
+else:
+
+    # possibly update PATH
+    os.environ['PATH'] = _maybe_prepend(os.environ.get('PATH', ''),
+                                        os.environ.get('BCBIOPATH', ''))
+    del _maybe_prepend
--- a/bcbio/setpath.py
+++ b/bcbio/setpath.py
@@ -4,57 +4,59 @@
 
 import os
 
-def _maybe_prepend(original, to_prepend):
-    """Prepend unique non-empty paths unless already present.
+def _prepend(original, to_prepend):
+    """Prepend paths in a string representing a list of paths to another.
 
-    original and to_prepend are expected to be strings corresponding to
+    original and to_prepend are expected to be strings representing
     os.pathsep-separated lists of filepaths.
 
-    If to_prepend is the empty sting, original is returned.
+    If to_prepend is None, original is returned.
+
+    The list of paths represented in the returned value consists of the first of
+    occurrences of each non-empty path in the list obtained by prepending the
+    paths in to_prepend to the paths in original.
 
     examples:
     # Unix
-    _maybe_prepend('/foo:/bar:/foo', '/foo:/baz:/baz') -> '/baz:/foo:/bar:/foo'
-    _maybe_prepend('/baz', '/foo::/bar:/foo')          -> '/foo:/bar:/baz'
-    _maybe_prepend('/foo:/bar:/foo', '')               -> '/foo:/bar:/foo'
+    _prepend('/b:/d:/a:/d', '/a:/b:/c:/a') -> '/a:/b:/c:/d'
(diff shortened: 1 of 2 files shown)


gberriz:
The two commits in this PR should be squashed.

chapmanb:
Gabriel;
Thanks much for the working on this, I appreciate you making this happen. A couple of initial thoughts:

- Could we make this a standard function call instead of happening via import? We generally try to avoid these kind of magic side effects to make what is happening very clear.
- We'll also need to add this to a couple of other places for when bcbio is run distributed and will be using a new process not deriving from the base one:

  - https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/distributed/runfn.py
  - https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/distributed/ipythontasks.py#L29

- I'm not sure how we envision this working with remote tasks with the environmental variable since it would require BCBIOPATH to be carried over. Maybe it would make sense to just swap over to doing this globally and calculating the correct path directly in the code to avoid the need for the extra variable. What do folks think?

Thanks again for starting this process.

gberriz:
On 10/25/2017 08:53 AM, Brad Chapman wrote:

>  *
>
>     Could we make this a standard function call instead of happening
>     via import? We generally try to avoid these kind of magic side
>     effects to make what is happening very clear.
>

Certainly.  I'm not very fond of magic myself.  My main reason for doing 
it the way I did it was to sidestep the question of where exactly in the 
client code the function is to be called, since I think that to answer 
this question one needs a greater familiarity with the bcbio codebase 
than I have.

Therefore, I've addressed these concerns through two PR's:

https://github.com/chapmanb/bcbio-nextgen/pull/2128
https://github.com/chapmanb/bcbio-nextgen/pull/2129

#2128 consists only of a stand-alone bcbio/setpath.py module, modified 
to get rid of the on-import side-effect.  By itself, the change entailed 
by this PR would have no effect at all on the behavior of the rest of 
the code.

#2129 essentially extends #2128 by modifying bcbio_nextgen.py to 
actually invoke the function defined in bcbio/setpath.py.

I've split the proposed changes over two PR's like this in case if my 
choice of where to invoke the prepend_bcbiopath function (which is what 
#2129 brings in) is not optimal, then you can just adopt #2128, and 
someone in the bcbio dev team can provide a more knowledgeable 
alternative to #2129.

>  *
>
>     We'll also need to add this to a couple of other places for when
>     bcbio is run distributed and will be using a new process not
>     deriving from the base one:
>
>       o https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/distributed/runfn.py
>         <https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_chapmanb_bcbio-2Dnextgen_blob_master_bcbio_distributed_runfn.py&d=DwMFaQ&c=WO-RGvefibhHBZq3fL85hQ&r=OO4mxmtiLsMtI-dWNh49fJe1In-YFyT1cY4fhd7Z4qo&m=S9iyS8jyn3KDliUQRb6KFz8LhMJBfu4U795CBKqnMgo&s=c-rv3u6f1-UGQRq-_s8RY6H3wMKbLgZIBIMwsOh6wVw&e=>
>       o https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/distributed/ipythontasks.py#L29
>         <https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_chapmanb_bcbio-2Dnextgen_blob_master_bcbio_distributed_ipythontasks.py-23L29&d=DwMFaQ&c=WO-RGvefibhHBZq3fL85hQ&r=OO4mxmtiLsMtI-dWNh49fJe1In-YFyT1cY4fhd7Z4qo&m=S9iyS8jyn3KDliUQRb6KFz8LhMJBfu4U795CBKqnMgo&s=U_5zx5inQ4AZhn-Cw72NvnBXEPULk6wtg14Mhj1hTI4&e=>
>  *
>
>     I'm not sure how we envision this working with remote tasks with
>     the environmental variable since it would require BCBIOPATH to be
>     carried over. Maybe it would make sense to just swap over to doing
>     this globally and calculating the correct path directly in the
>     code to avoid the need for the extra variable. What do folks think?
>

I'm less clear on how to address your last two points, since I'm not 
familiar with the distributed and remote features you mention.  I have a 
few questions/thoughts about them, but I think it'd be more efficient 
for us to go over them in an in-person meeting.  I can come by your 
office any time today before 3pm, or tomorrow except for 1-2pm.

Best,

G.

gberriz:
On 10/25/2017 08:53 AM, Brad Chapman wrote:
> * Could we make this a standard function call instead of happening via import?  We generally try to avoid these kind of magic side effects to make what is happening very clear.

Certainly.  I'm not very fond of magic myself.  My main reason for doing 
it the way I did it was to sidestep the question of where exactly in the 
client code the function is to be called, since I think that to answer 
this question one needs a greater familiarity with the bcbio codebase 
than I have.

Therefore, I've addressed these concerns through two PR's:

https://github.com/chapmanb/bcbio-nextgen/pull/2128
https://github.com/chapmanb/bcbio-nextgen/pull/2129

#2128 consists only of a stand-alone `bcbio/setpath.py` module, modified 
to get rid of the on-import side-effect.  By itself, the change entailed 
by this PR would have no effect at all on the behavior of the rest of 
the code.

#2129 essentially extends #2128 by modifying `bcbio_nextgen.py` to 
actually invoke the function defined in bcbio/setpath.py.

I've split the proposed changes over two PR's like this in case if my 
choice of where to invoke the `prepend_bcbiopath` function (which is what 
#2129 brings in) is not optimal, then you can just adopt #2128, and 
someone in the bcbio dev team can provide a more knowledgeable 
alternative to #2129.

> * We'll also need to add this to a couple of other places for when bcbio is run distributed and will be using a new process not deriving from the base one:
>
>   * https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/distributed/runfn.py
>   * https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/distributed/ipythontasks.py#L29
>
> * I'm not sure how we envision this working with remote tasks with the environmental variable since it would require BCBIOPATH to be carried over. Maybe it would make sense to just swap over to doing this globally and calculating the correct path directly in the code to avoid the need for the extra variable. What do folks think?

I'm less clear on how to address your last two points, since I'm not 
familiar with the distributed and remote features you mention.  I have a 
few questions/thoughts about them, but I think it'd be more efficient 
for us to go over them in an in-person meeting.  I can come by your 
office any time today before 3pm, or tomorrow except for 1-2pm.

Best,

G.

chapmanb:
Gabriel;
Thanks for the discussion and the 2 PRs. The bcbio tests are currently broken due to some unrelated software changes so the CI failures aren't due to your adjustments. I hope to have those fixed later today.

I won't be at Longwood today/tomorrow for an in person discussion but can take a crack at the improvements I mention in the last paragraph after merging your changes. Thanks again for these and I'll work on getting them integrated.

gberriz:
Superseded by #2128 or #2129.

PR #2126, opened 2017-10-26
Coverage: add thresholds

vladsaveliev:
When `coverage` BED file provided, calculate regions completeness for a set of depth thresholds. It's similar to what was done using sambamba before we migrated to mosdepth. Requires the new mosdepth [release](https://github.com/brentp/mosdepth/releases/tag/v0.2.0), which is already added to bioconda. 

Note the difference with samamba. It reported percentages:
```
chrom  chromStart  chromEnd  F3  readCount  meanCoverage  percentage1  percentage5  percentage10  sampleName
chr21  9483184     9483680   .   225        33.7641       100          98.7903      73.9919       sample
```
whereas mosdepth reports the region lengths in base pairs:
```
#chrom  start   end     region  1X      5X      10X
chr21   9483184 9483680 .       496     493     357
```
I think it's fine to stick to the lengths because percentages can be calculated out of that, however the other way around is not always possible.

This is the list of coverage-qc-related files we save into `qc/coverage`:
```
*-coverage.mosdepth.dist.txt         # total and per-chromosome coverage distribution
*-coverage.regions.bed.gz            # average depths in each region
*-coverage.thresholds.bed.gz         # the new file with coverage thresholds for each region
```
I removed the `_caluclate_percentiles` function call in `coverage_region_detailed_stats` which used to generate `*_bcbio_coverage_avg.txt` because it duplicates `*-coverage.mosdepth.dist.txt`. So this also requires changes in MultiQC_bcbio https://github.com/MultiQC/MultiQC_bcbio/pull/16 which uses that file to make the coverage distribution plot. Note that we are also no longer generate `*_bcbio_coverage.txt` file (used for the "regions completeness" plot), though I never used neither the file nor the plot so unless we will get requests, I'd put it on hold.

For all coverage calculations in bcbio so far, only one mosdepth run is required per sample per BED file, which is great.


chapmanb:
Vlad -- thanks for incorporating this so quickly. I'm happy to have the threshold functionality back in. The CI failures are due to an old version of mosdepth in the Docker container. I'll work on updating the multiqc package with your other PR and re-build this so we can get tests working cleanly. Thanks again.

biocyberman:
Just to discuss, not to oppose this PR. I would prefer sambamba output, becuase it does not require an extra step to see how much an interval is covered. This number is not much far off when we have 4 or 5 digit precision: `"number of covered bases = covered percentage x interval lenght/100"`. When it is not possible this way - maybe because interval length is not there - so then the number of bases like in `mosdepth` output becomes even less meaningful than percentage. But this is not a big deal anyway, I can live with doing one extra step to calculate percentage. 

Talking about coverage, clinicians are very interested in having exon-level coverages as well, which `chanjo` tries to achieve with some success. How is bcbio-nextgen is doing about this Brad @chapmanb ?

chapmanb:
Vang;
Thanks for the discussion about this. Unfortunately sambamba depth had intermittent segfaults and failures for many users so we've moved away from using it. mosdepth is a great integrated tool with improved speed, and is also centralized in the new bcbio coverage calculation infrastructure, so I hope it will overall be a positive transition now that we're feature complete with previous usage thanks to Vlad.

We explored chanjo integration in the past in bcbio and liked the tool but struggled with a way to use it in an automated way and swapped over to using `coverage` with MultiQC reporting:

https://bcbio-nextgen.readthedocs.io/en/latest/contents/configuration.html#analysis-regions

would having a BED file of exons there provide the sort of statistics you'd need? Thanks again for the discussion.

biocyberman:
Thanks for clarification, Brad. With a BED file of exons, `coverage` report from MultiQC is useful for this purpose. However, it does not help if one wishes to see reports for a gene across samples or across multiple runs.

chapmanb:
Vang -- thanks for the feedback. Could bcbio output coverage in a more useful way to feed into custom downstream use of Chanjo or other tools? You're exactly right in that the analysis based approach we take in bcbio is not compatible with multi-run comparisons but we aim to produce outputs that are relatively easy to merge and use. If you have a workflow that we could make easier happy to explore putting together a plan for it. Thanks again for the discussion.

PR #2128, opened 2017-10-26
bcbio/setpath.py: initial commit

gberriz:
The code in this PR addresses the first point in https://github.com/chapmanb/bcbio-nextgen/pull/2124#issuecomment-339320006 .

In this version of the `bcbio/setpath.py`, the modification to `PATH` can happen only through a function call, rather than through a side-effect of `import`.  (I will include modifications involving the actual invocations of this function in a separate PR.)


chapmanb:
Merged in #2129

PR #2129, opened 2017-10-26
Invoke prepend bcbiopath

gberriz:
The code in this PR addresses the first point in https://github.com/chapmanb/bcbio-nextgen/pull/2124#issuecomment-339320006 .

(NB: this PR extends #2128.)

### DIFF
--- a/bcbio/setpath.py
+++ b/bcbio/setpath.py
@@ -1,0 +1,65 @@
+"""Update the PATH environment variable to reflect the value of BCBIOPATH.
+
+"""
+
+import os
+
+def _prepend(original, to_prepend):
+    """Prepend paths in a string representing a list of paths to another.
+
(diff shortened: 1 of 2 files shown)


gberriz:
I don't understand why the commits in this PR failed, but those in #2124 succeeded, since, in content, they are essentially the same.

chapmanb:
Gabriel -- thank you again for this and sorry for the delay in merging; we wanted to push out the 1.0.6 release so we could test this out. I'll add in some additional commits to add this automatically for parallel runs as well. Thanks again for doing this.

chapmanb:
Gabriel -- I checked in a slight extension of this to apply to parallel runs and also use the internal bcbio anaconda path if BCBIOPATH is not set. Please let us know if you run into any problems during modules testing. Thank you again for putting this together.

Issue #2122, opened 2017-10-24
Use Warning instead of Exception at data upload ?


biocyberman:
The following exception was thrown because `/proj11/work/structural/Son1/cnvkit/raw/Son1-sort-recal-mybatch1.cnr` does not exist (on why it does not exist, I don't know yet).

```
[2017-10-24T08:52Z] [INFO   ]         multiqc : MultiQC complete
Traceback (most recent call last):
  File "/bcbio/bin/bcbio_nextgen.py", line 234, in <module>
    main(**kwargs)
  File "/bcbio/bin/bcbio_nextgen.py", line 43, in main
    run_main(**kwargs)
  File "/bcbio/anaconda/lib/python2.7/site-packages/bcbio/pipeline/main.py", line 42, in run_main
    fc_dir, run_info_yaml)
  File "/bcbio/anaconda/lib/python2.7/site-packages/bcbio/pipeline/main.py", line 86, in _run_toplevel
    for xs in pipeline(config, run_info_yaml, parallel, dirs, samples):
  File "/bcbio/anaconda/lib/python2.7/site-packages/bcbio/pipeline/main.py", line 182, in variant2pipeline
    samples = qcsummary.generate_parallel(samples, run_parallel)
  File "/bcbio/anaconda/lib/python2.7/site-packages/bcbio/pipeline/qcsummary.py", line 38, in generate_parallel
    samples = run_parallel("multiqc_summary", [samples])
  File "/bcbio/anaconda/lib/python2.7/site-packages/bcbio/distributed/multi.py", line 28, in run_parallel
    return run_multicore(fn, items, config, parallel=parallel)
  File "/bcbio/anaconda/lib/python2.7/site-packages/bcbio/distributed/multi.py", line 86, in run_multicore
    for data in joblib.Parallel(parallel["num_jobs"], batch_size=1)(joblib.delayed(fn)(x) for x in items):
  File "/bcbio/anaconda/lib/python2.7/site-packages/joblib/parallel.py", line 804, in __call__
    while self.dispatch_one_batch(iterator):
  File "/bcbio/anaconda/lib/python2.7/site-packages/joblib/parallel.py", line 662, in dispatch_one_batch
    self._dispatch(tasks)
  File "/bcbio/anaconda/lib/python2.7/site-packages/joblib/parallel.py", line 570, in _dispatch
    job = ImmediateComputeBatch(batch)
[...]

PR #2132, opened 2017-10-30
Coverage thresholds checks

vladsaveliev:
To address #2122 - rerun mosdepth if the coverage thresholds file is missing; if mosdepth is older, just skip processing thresholds.

### DIFF
--- a/bcbio/variation/coverage.py
+++ b/bcbio/variation/coverage.py
@@ -3,18 +3,17 @@
 Provides estimates of coverage intervals based on callable regions
 """
 import collections
-import csv
 import itertools
 import os
 import shutil
+from subprocess import check_output
+from distutils.version import LooseVersion
 import yaml
-
 import pybedtools
 import numpy as np
 import pysam
 import toolz as tz
 
-from bcbio.utils import (append_stem, copy_plus)
 from bcbio import bam, utils
 from bcbio.bam import ref, readstats
 from bcbio.distributed.transaction import file_transaction
@@ -208,6 +207,7 @@ def run_mosdepth(data, target_name, bed_file, per_base=False, quantize=None, thr
     bam_file = dd.get_align_bam(data) or dd.get_work_bam(data)
     work_dir = utils.safe_makedir(os.path.join(dd.get_work_dir(data), "coverage", dd.get_sample_name(data)))
     prefix = os.path.join(work_dir, "%s-%s" % (dd.get_sample_name(data), target_name))
+    thresholds = thresholds and LooseVersion(check_output(["mosdepth", "--version"]).split()[1].strip()) >= LooseVersion("0.2.0")
     out = MosdepthCov("%s.mosdepth.dist.txt" % prefix,
                       ("%s.per-base.bed.gz" % prefix) if per_base else None,
                       ("%s.regions.bed.gz" % prefix) if bed_file else None,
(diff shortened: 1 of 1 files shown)


bcbio/variation/coverage.py
```diff
@@ -208,12 +207,17 @@ def run_mosdepth(data, target_name, bed_file, per_base=False, quantize=None, thr
     bam_file = dd.get_align_bam(data) or dd.get_work_bam(data)
     work_dir = utils.safe_makedir(os.path.join(dd.get_work_dir(data), "coverage", dd.get_sample_name(data)))
     prefix = os.path.join(work_dir, "%s-%s" % (dd.get_sample_name(data), target_name))
+    thresholds = thresholds and LooseVersion(check_output(["mosdepth", "--version"]).split()[1].strip()) >= LooseVersion("0.2.0")
```
chapmanb: We can assume that the user will have an up to date mosdepth so always apply the thresholds argument. The case we want to catch is when mosdepth has already been run with an older version (so won't have the thresholds file).

bcbio/variation/coverage.py
```diff
@@ -208,12 +207,17 @@ def run_mosdepth(data, target_name, bed_file, per_base=False, quantize=None, thr
...
                       ("%s.per-base.bed.gz" % prefix) if per_base else None,
                       ("%s.regions.bed.gz" % prefix) if bed_file else None,
                       ("%s.quantized.bed.gz" % prefix) if quantize else None,
                       ("%s.thresholds.bed.gz" % prefix) if thresholds else None)
-    if not utils.file_uptodate(out.dist, bam_file):
+    if not utils.file_uptodate(out.dist, bam_file) \
+            or (per_base and not utils.file_uptodate(out.per_base, bam_file)) \
+            or (bed_file and not utils.file_uptodate(out.regions, bam_file)) \
+            or (quantize and not utils.file_uptodate(out.quantize, bam_file)) \
+            or (thresholds and not utils.file_uptodate(out.thresholds, bam_file)):
```
chapmanb: I'd skip all these extra checks here. If the user has run it previously we don't want to re-run, but later just skip if the thresholds file isn't present.

bcbio/variation/coverage.py
```diff
@@ -226,9 +230,10 @@ def run_mosdepth(data, target_name, bed_file, per_base=False, quantize=None, thr
                 quant_export += " && "
             else:
                 quant_arg, quant_export = "", ""
-            thresholds = "-T " + ",".join([str(t) for t in thresholds]) if thresholds else ""
+
+            thresholds_cmdl = "-T " + ",".join([str(t) for t in thresholds]) if out.thresholds else ""
```
chapmanb: This is causing the test failure (https://travis-ci.org/chapmanb/bcbio-nextgen/builds/294676740#L4918) but we can revert back have this as long as we have thresholds set (assume mosdepth 0.2.0 or better installed).

bcbio/variation/coverage.py
```diff
@@ -252,12 +257,12 @@ def coverage_region_detailed_stats(target_name, bed_file, data, out_dir):
         ready_depth = tz.get_in(["depth", target_name], data)
         cov_file = ready_depth["regions"]
         dist_file = ready_depth["dist"]
-        thresholds_file = ready_depth["thresholds"]
+        thresholds_file = ready_depth.get("thresholds")
```
chapmanb: If you add an existence check here and set thresholds_file to None if missing I think that should capture what we need.

chapmanb:
Vlad -- thanks much for tackling this issue (from #2122 discussion). I left some throughts: my approach would be to always assume mosdepth 0.2.0 or greater if running from scratch and then check afterwards and only return the thresholds file if present. Thanks again for looking at this.

vladsaveliev:
Passing now!

chapmanb:
Vlad -- thanks for all this work, much appreciated.

PR #2147, opened 2017-11-09
Update pandas sort syntax for integration/rnaseq/test_ericscript.py

biocyberman:
Although a tiny fix, it helps to get all test in ` integration/rnaseq/test_ericscript.py` finish successfully :) 
The next thing is to get `CheckDB.R`  of the original `ericrscipt`  to work correctly with bcbio_nextgen bwa version. The file is in here: samtools0/share/ericscript-0.5.5-1/lib/R/CheckDB.R

### DIFF
--- a/tests/integration/rnaseq/test_ericscript.py
+++ b/tests/integration/rnaseq/test_ericscript.py
@@ -135,6 +135,6 @@ def _load_result_file(fname):
         'fusiontype'
     ]
     sort_by = ['EnsemblGene1', 'EnsemblGene2']
-    df = pd.read_csv(fname, sep='\t')[columns_to_keep].sort(sort_by)
+    df = pd.read_csv(fname, sep='\t')[columns_to_keep].sort_values(by=sort_by)
     df.index = range(len(df))
     return df


chapmanb:
Vang -- thanks much for the fix. It looks like the tests failed for an unrelated Docker reason.

For the CheckDB.R problem, I think we should do change line 36 from:
```
  if (curr.version.bwa != prev.version.bwa) {
```
to
```
  if (FALSE) {
```
so it never does an inplace update. If this works for you we can incorporate into the bioconda recipe.

Thanks again.

biocyberman:
@chapmanb The whole R script as a whole doesn't do much, and I don't know the reason to write it in R, probably for better portability, or to work with other R scripts. Anyway, what works for me to to comment out line 31 - line 58. You can make a patch to comment the lines in the bioconda recipe.

chapmanb:
Vang -- thanks much, I pushed this fix to the bioconda recipe so a tools update should get the fixed version. Please let me know if you run into any other issues at all. We're planning to use this on a bigger project for more testing and appreciate you working through any issues you're hitting. Looking forward to hearing how it performs.

PR #2152, opened 2017-11-13
MultiQC secondary files: add multiqc_data.json, remove globs

vladsaveliev:
Following the discussion in Slack, adding `multiqc_data.json` and cleaning up the secondary file list to get rid of globs, so it makes it more suitable for CWL definitions. Not adding changes to `cwl/defs.py` yet as I still don't have a test set up for that and don't want to break anything.

Also fixing work-rooted paths in `multiqc_data.json`, similarly to those in `list_files.txt`, to make MegaQC function out of just `final` directory.


chapmanb:
Vlad -- thanks so much. I think the test issue was primarily related to not having MultiQC 1.3 in the latest Docker build. I'll re-build the latest to get both this and 1.3 included. Thanks again.

Issue #2154, opened 2017-11-13
"germline" present in multiple tumor samples causes multiple QC runs of the normal (?)


lbeltrame:
Note the (?) in the title, because I'm not sure what is the actual cause. I noticed that one of my normal samples was run repeatedly over QC (as up as 5-6 times): this included downsampling, mapped reads with samtools and FastQC runs (at a point some failed because the file they were using got overwritten).

The YAML template specified `germline` variant calling for all samples, out of simplicity because I know that the setting is ignored in tumor samples. That's my (current) only clue, but I can't test easily the same configuration without it (very deep reads, long time processing).

lbeltrame:
I've noticed this also with other samples: QC failed because two nodes ran the same QC (`idxstats`), one finishing earlier than the other and thus causing an exception for `shutil` (because the file already existed).

chapmanb:
Luca -- thanks for the report. I'm cc'ing in @vladsaveliev since he worked on germline QC most recently and would have the best understanding of potential issues. Can you describe your sample YAML so we can try to reproduce? Is this a normal shared amongst multiple tumors? Ideally we'd try to replicate the behavior on a small test to help us diagnose. Any error messages or tracebacks you saved from the race condition would also be helpful to pinpoint issues. Thanks much.

vladsaveliev:
Thanks for the report!

I created a PR that will hopefully fix it. The issue was that during one of the stages, bcbio was adding a copy of normal into the processing list for each batch the normal belongs. So I fixed it to regroup samples back as before this stage, keeping only one normal sample.

Vlad

chapmanb:
Luca -- this should be fixed in latest development with #2157. Please let us know if you run into any other issues and thank you again for letting us know about the issue.

PR #2157, opened 2017-11-14
Regroup back samples after processing in heterogeneity module

vladsaveliev:
In the `heterogeneity` module, "normal" samples get duplicated when shared between multiple batches, leading to duplicated QC work downstream. This PR regroups samples after `heterogeneity` processing, to avoid of redundant normal sample copies. Resolves duplicated QC runs for #2154

### DIFF
--- a/bcbio/heterogeneity/__init__.py
+++ b/bcbio/heterogeneity/__init__.py
@@ -8,6 +8,7 @@ especially in complex cancer samples.
 from __future__ import print_function
 import collections
 
+from bcbio import utils
 from bcbio.heterogeneity import bubbletree, phylowgs, theta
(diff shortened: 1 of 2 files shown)


vladsaveliev:
Another way to resolve this would have been to keep track of normal samples, and for all duplicates, clean up `["config"]["algorithm"]["qc"] - that code could be placed into `qc` module. But I'm thinking this solution is more elegant. Hope it's not breaking anything.

chapmanb:
Vlad, thanks much for catching this and for the quick fix. I definitely prefer having the samples be consistent in and out of different steps so +1 for putting the fix directly in heterogeneity calculations. Thank you again.

PR #2161, opened 2017-11-17
Fix typos.

cure:


chapmanb:
Thanks Ward for catching this. I have no idea how I made the same typo twice, glad to have it fixed up.

PR #2169, opened 2017-11-23
Normalize calls before ensemble

vladsaveliev:
Adds normalization as an intermediate step for ensemble calling:
- split multiallelic variants,
- left-aligns and normalize indels,
- decompose biallelic MNP into single SNP.
- after calling ensemble VCF, re-add snpEff annotations.

Thanks for the discussions at https://github.com/chapmanb/bcbio-nextgen/issues/2149

`multiallelic` module was renamed to `normalize` as it contains now 3 types of normalization. Extended docs to reflect that.

Also, as a minor improvement to the normalization command, moved `bcftools view -f 'PASS,.'` command to the beginning to avoid doing redundant work on rejected calls.


vladsaveliev:
Here is the comparison before/after for a GiaB somatic-like mixture:
screen shot 2017-11-23 at 13 41 06

chapmanb:
Vlad -- thanks so much for this work, this is brilliant. It's nice to see the improvement in sensitivity with only a small decrease in specificity. Much appreciated.

PR #2183, opened 2017-12-15
fix print statements so that there is no error in python3

etolpygo:


chapmanb:
Thank you for cleaning this up. Much appreciated.

PR #2184, opened 2017-12-15
Accept non-integer memory amounts (e.g. 7.6G).

cbrueffer:


chapmanb:
Christian -- thanks for this. I typically use Mb values when getting into sub-Gb specifications but happy to have support for both. Much appreciated.

PR #2186, opened 2017-12-16
Add Break Point Inspector for Manta calls filtering

vladsaveliev:
Hi Brad,

Adding Hartwig's BPI for Manta call filtering: https://github.com/hartwigmedical/hmftools/tree/master/break-point-inspector
Can be turned on with `tools_on: break-point-inspector`.

BPI significantly reduces the number of reported PASSed calls, 2-3 times in WGS samples that we experimented with, and we didn't spot any loss in sensitivity so far (though never actually benchmarked). Good to have it optional until we perform evaluation on a truth set, or Illumina guys consider adding it into Manta.

Needs this Bioconda PR to be accepted: https://github.com/bioconda/bioconda-recipes/pull/7061

Vlad

### DIFF
--- a/bcbio/cwl/defs.py
+++ b/bcbio/cwl/defs.py
@@ -375,7 +375,7 @@ def _variant_sv(checkpoints):
                             cwlout(["sv", "vrn_file"], ["File", "null"], [".tbi"]),
                             cwlout("inherit")])],
             "bcbio-vc", ["bedtools", "cnvkit", "delly", "extract-sv-reads",
-                         "lumpy-sv", "manta", "mosdepth", "samtools",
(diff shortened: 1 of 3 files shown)


chapmanb:
Vlad -- thank you for this contribution, glad to have it in. After initial testing, do you think it would make sense to have this turned on automatically for tumor/normal runs with manta? If this works, we could then remove the `tools_on` requirement. Thanks again.

vladsaveliev:
Brad, thanks for accepting. Yes, I think it can be turned on automatically (at least we are going to keep it on in UMCCR). It's soft filtering, so all original calls are kept.

PR #2192, opened 2017-12-19
In postprocess_alignment, deduplicate if requested. Fixes #2188

### LINKED ISSUE #2188: variant2pipeline workflow not marking duplicates?
Hi everyone,

the variant2pipeline workflow does not mark duplicates after alignment. I noticed this with bcbio-nextgen 1.0.2, and verified it with 1.0.6.

The config template I use is as follows (duplication marking is explicitly enabled):

```
details:
  - analysis: variant
    genome_build: hg38
    algorithm:
      # alignment
      platform: Illumina
      strandedness: firststrand
      aligner: hisat2
      mark_duplicates: true
      trim_reads: false
      recalibrate: false
      realign: false
      # variant calling
      variantcaller: [vardict]
      remove_lcr: false
      min_allele_fraction: 2
      effects: false
```

The workflow calls ```pipeline.sample.postprocess_alignment()```, where I would have expected duplicate marking to take place, but it's currently missing.

Is this intentional, or did I miss something obvious?

cbrueffer:
Proposed fix for issue #2188 

Not sure this is ideal; e.g. when the workflow input was a BAM file, duplicates are already marked as part of the BAM cleaning process.

### DIFF
--- a/bcbio/pipeline/sample.py
+++ b/bcbio/pipeline/sample.py
@@ -226,6 +226,7 @@ def postprocess_alignment(data):
         bam_file_ready = os.path.join(out_dir, os.path.basename(bam_file))
         if not utils.file_exists(bam_file_ready):
             utils.symlink_plus(bam_file, bam_file_ready)
+        bam_file_ready = postalign.dedup_bam(bam_file_ready, data)
         bam.index(bam_file_ready, data["config"])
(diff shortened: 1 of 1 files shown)


chapmanb:
Christian -- the right fix is to support streamed de-duplication with hisat2 as this would end up doing double de-duplication for other callers. I'll close this and continue discussion in #2188. Thanks so much for reporting and looking at this.

PR #2217, opened 2018-01-12
made cosmic ref file name more generic, added pointer for hg38, docs

matthdsm:
fixes https://github.com/chapmanb/cloudbiolinux/issues/262
Docs may need some more work to fit with the rest.
Ping @lemdcock

Cheers
M


chapmanb:
Matthias and Laurenz;
Thanks much for taking these on. I appreciate the improvements.

PR #2218, opened 2018-01-15
Fix manta+bpi for "germline" samples

vladsaveliev:
For the samples with `germline` phenotype, `paired` object is `None`, making this piece of code crash. Adding an extra fix. Thanks @ohofmann for spotting the bug.


chapmanb:
Vlad -- thank you for debugging this and the quick fix.

PR #2221, opened 2018-01-15
Add a prefix for the variant regions to avoid filename output clobbering

geetduggal:
@chapmanb Not sure if other operations depend on the name in particular, but this could avoid the filename output clobbering that may be happening when cwltool is running.  Here is a proposed PR if that's the case.


chapmanb:
Thanks Geet, this is a nice workaround. We aren't dependent on explicit file names so happy to prefix if that helps with cwltool output manipulation.

Issue #2267, opened 2018-02-14
Clarify CNVkit usage with regards to sample gender


lbeltrame:
cc @LMannarino

In some of our recent analyses there has been some confusion on the output of CNVkit with regards to the use of sex chromosomes, as outlined in https://github.com/etal/cnvkit/issues/316. As far as I could dig in the bcbio code, `-y / --male-reference` for male samples is only used in diagrams, but not anywhere else (feel free to correct me if I got it wrong, though). 
According to the aforementioned issue, it should not be used if it had not been used elsewhere (`call` and `background` included).
The reason is that we're getting consistently reports of loss on chrX in some of our analyses, and we don't know if it's an artifact or a real alteration.

lbeltrame:
Poking also @etal for further comments here.

etal:
Did my comment in the linked ticket clarify the problem or reveal a bug? (I edited my comment there after a hasty initial response, but the automatic e-mail response you saw would have been fairly unhelpful.)

If `-y` wasn't used to construct the reference, it shouldn't be used anywhere else, including `diagram`. The development version of the documentation provides more detail than the current stable version on both sample/reference sex and `diagram`:
- https://cnvkit.readthedocs.io/en/latest/sex.html
- https://cnvkit.readthedocs.io/en/latest/plots.html#diagram

In general, for a male sample normalized to a reference with the default ploidies, chrX will have a log2 value around -1, which you can confirm in the .cnr/.cns files, a `scatter` plot, and the output of the `sex` and `gainloss` commands. In the output of `call`, you'll also see in the `cn` column that the chrX integer copy number is 1. The development version of `diagram` should figure out that the sample is male and display chrX as copy-number-neutral (white), unless you add the `-y` flag, in which case it will take the -1 log2 value literally and display chrX in blue -- I don't remember whether that's also the behavior in the stable release; in earlier versions it wouldn't figure out that male samples normally have haploid chrX, so chrX would always be blue.

lbeltrame:
> Did my comment in the linked ticket clarify the problem or reveal a bug?

It *may* have uncovered a bug. I see that `-y` isn't used anywhere but `diagram`, so I assume the plots done by diagram in bcbio are inaccurate for the sex chromosomes in the case of male samples.

etal:
Probably an oversight -- I did a code review in another bcbio issue related to this and I might have missed the `diagram` section at the time. I'll look for it in the current code.

chapmanb:
[...]

PR #2268, opened 2018-02-14
cnvkit diagram: Fix chrX display for male samples

etal:
Per issue #2267 and following up on f008e6341d207867ec0a7d4ff3d75a27269f92ad, the `--male-reference` isn't needed here, but the `--sample-sex` option can still be useful, matching the `cnvkit.py call` usage.


chapmanb:
Eric -- thanks so much for the review and quick fix.

PR #2276, opened 2018-02-18
Apply min_allele_fraction for Strelka2 and MuTect2 calls

vladsaveliev:
Currently, `min_allele_fraction` is applied to VarDict, Freebayes, Varscan, and Scalpel, but not to Mutect and Strelka2 calls. In this PR, adding support for that for the latter callers.

For Strelka2, calculating and populating `FORMAT/AF` field, which is derived from existing `AD` (germline) or `*U` (somatic snps) or `TIR` (somatic indels) FORMAT fields that represent allelic depths.

Low-AF variants are soft-filtered by adding `MinAF` into `FILTER`. Let me know if it makes sense to rather hard-filter.

Also, I figured that's not really possible to pipe filtering from the existing variant calling commands, because we can't reliably know in advance the position of the tumor sample column in VCF, so need to have VCF file saved before filtering. But that doesn't seem to add any noticeable overhead.

### DIFF
--- a/bcbio/variation/germline.py
+++ b/bcbio/variation/germline.py
@@ -97,7 +97,7 @@ def _remove_prioritization(in_file, data):
     return out_file
 
 def _update_prioritization_filters(rec):
-    rec = _remove_filter(rec, "LowPriority")
+    rec = vcfutils.cyvcf_remove_filter(rec, "LowPriority")
     return _update_germline_filters(rec)
 
 def _extract_germline(in_file, data):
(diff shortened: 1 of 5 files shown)


chapmanb:
Vlad -- thanks so much for adding and normalize allele fraction calculations across all these callers. I'm definitely agreed about soft-filtering so that folks can decide to adjust afterwards as needed in downstream pipelines. Thanks again.

vladsaveliev:
Thanks for merging, and for a fix for reruns!

PR #2279, opened 2018-02-20
Strelka2 and Mutect2 MinAF check: use np.all to check multiallelic

vladsaveliev:
Minor fix for the MinAF check. For multiallelic variants, we want to reject only those where all alleles are below the threshold.


chapmanb:
Vlad -- thanks much for the additional fixes.

PR #2297, opened 2018-03-01
A fix for SV validation plot

naumenko-sa:
When I ran SV validation for svcaller: [lumpy, manta, cnvkit, metasv, wham], the sv-validate.png plot displayed caller labels incorrectly. This is a fix of that.


chapmanb:
Thank you for the plot cleanup. Much appreciated.

PR #2380, opened 2018-04-27
Add `coverage_qc` to `tools_off`

vladsaveliev:
This allows to skip running `samtools-stats`


chapmanb:
Brilliant, thanks so much for the enhancement Vlad.

PR #2381, opened 2018-04-27
Viral content: calculate coverage completeness instead of read counts

vladsaveliev:
In viral stats table, bcbio reports the number of reads supporting each GDC viral sequence. This metric doesn't take into account the size of the virus; And also, in some tests I ran into viruses that contained a small repeat that attracted thousands of reads, resulting in one colossal coverage peak, with 0 coverage all along the rest of the viral sequence.

To account for those issues, I suggest to calculate coverage completeness - e.g. % viral sequence covered at 5x - and rate the viruses based on this metric instead of read count.

Will submit a corresponding MultiQC_bcbio soon.

### DIFF
--- a/bcbio/qc/viral.py
+++ b/bcbio/qc/viral.py
@@ -13,8 +13,12 @@ from bcbio.provenance import do
 from bcbio.variation import vcfutils
 
 def run(bam_file, data, out_dir):
-    """Run viral QC analysis.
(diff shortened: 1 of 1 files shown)


chapmanb:
Nice one, thanks for making the viral QC calculations more useful. I'll work on updating the MultiQC bcbio conda module as well to make use of this.

PR #2407, opened 2018-05-29
cromwell: torque qsub fix

matthdsm:
removed invalid flags from cromwell torque job submission config

### DIFF
--- a/bcbio/cwl/hpc.py
+++ b/bcbio/cwl/hpc.py
@@ -208,9 +208,9 @@ HPC_CONFIGS = {
         %(cwl_attrs)s
         \"\"\"
         submit = \"\"\"
-        qsub -V -l wd -N ${job_name} -o ${out} -e ${err} -q ${queue} -l walltime=${walltime} \
+        qsub -V -N ${job_name} -o ${out} -e ${err} -q ${queue} -l walltime=${walltime} \
         %(cpu_and_mem)s \
-        -- /usr/bin/env bash ${script}
+        ${script}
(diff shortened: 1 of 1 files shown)


matthdsm:
Hi Brad,

Sorry about the mistake, The changes are supposed to be in the torque section.

As far as I know, torque has no way of setting the shell from the qsub command. However, I saw the submission scripts have a correct bash shebang, I think that should be enough for torque.

as an alternative to the `-l wd` flag you could use `-d "workdir"`. I suppose that's the torque equivalent.

for more flags etc, see http://docs.adaptivecomputing.com/torque/archive/4-1-1/Content/topics/commands/qsub.htm

Cheers
M

matthdsm:
I fixed the args and restored the PBSpro settings.

chapmanb:
Brilliant, thanks Matthias. Are we able to keep the `/usr/bin/env bash` invocation once we fix the `-l` to `-d` argument? I'd like to keep this in, if you can confirm it works, so that we're consistent in how we prepare these scripts between different schedulers. Thanks again for helping with this.

matthdsm:
The `-S`flag may be used the set the shell for a job, but the qsub command can't handle the whitespace in `/usr/bin/env bash`. If you're okay with using `/bin/bash` I suppose we're good to go.

M

chapmanb:
Thanks, if the `/usr/bin/env` invocation doesn't work happy to have the simplified version. Thank you again for testing and the fix.

PR #2411, opened 2018-06-06
CWL: cromwell torque "check-alive" fix

matthdsm:
torque `qstat` has no `-j` flag, resulting in faulty tracking of job status.


chapmanb:
Thanks much for testing this and for the fix. Hopefully this resolves the problem with Cromwell not realizing tasks had died you found earlier. I appreciate all the Torque testing.

matthdsm:
With pleasure! Thank you for the awesome work on bcbio.

M

PR #2428, opened 2018-06-29
Clarifying the OS expectation for the mentioned bcbio-vm install commands

skanwal:
Hopefully this will be heads-up if someone tries to run the install commands on MAC or any other platform


chapmanb:
Thanks so much for the improved documentation.

PR #2432, opened 2018-07-03
Atropos polyX trimming: disable anchored mode to allow variable length polyX stretches in 3' ends

vladsaveliev:
The dollar sign in the -a atropos options (`-a 'A{200}$' -a 'C{200}$' -a 'G{200}$' -a 'T{200}$'` and the corresponding `-A` options) require the adapter to be _exactly_ 200 bases long in the read 3' end. That's not the behaviour we want: we rather want it to be of _any_ length between 8 bases and 200 bases (as soon as it in the 3' end, and not in the middle of the read!). Removing $ corrects atropos behaviour to exactly the one we want:

```
test.fq:
@read1
GGGGGGGGGGGAAAAAAAAAAAAAAAAAAAAAA
```

```
atropos trim -a 'A{200}$' --overlap 8 --no-default-adapters -se test.fq
output:
@read1
GGGGGGGGGGGAAAAAAAAAAAAAAAAAAAAAA
```

```
atropos trim -a 'A{200}' --overlap 8 --no-default-adapters -se test.fq
output:
@read1
GGGGGGGGGGG
```

### DIFF
--- a/bcbio/bam/trim.py
+++ b/bcbio/bam/trim.py
@@ -70,7 +70,7 @@ def _atropos_trim(fastq_files, adapters, out_dir, data):
                 tx_out2 = tx_out[2]
             # polyX trimming, anchored to the 3' ends of reads
             if "polyx" in dd.get_adapters(data):
-                adapters += ["A{200}$", "C{200}$", "G{200}$", "T{200}$"]
+                adapters += ["A{200}", "C{200}", "G{200}", "T{200}"]
             adapters_args = " ".join(["-a '%s'" % a for a in adapters])
             adapters_args += " --overlap 8"  # Avoid very short internal matches (default is 3)
             adapters_args += " --no-default-adapters --no-cache-adapters"  # Prevent GitHub queries and saving pickles


vladsaveliev:
One minor thing to note. Atropos has a default allowed error rate in the adapter sequence of 10%. Because of that, in the example below not only the read1 will be trimmed, but read2 and read3 as well (but not read4):

```
@read1
GGGGGGGGGGGAAAAAAAAAAAAAAAAAAAAAA
+
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
@read2
GGGGGGGGGGGAAAAAAAAGAAAAAAACAAAAA
+
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
@read3
GGGGGGGGGGGAAAAAAAAAAAAAAAAAAAAGC
+
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
@read4
GGGGGGGGGGGAAAAAAAAAAAAAAAAAAAAGCG
+
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```

```
atropos trim -a 'A{200}' --overlap 8 --no-default-adapters -se test.fq

@read1
GGGGGGGGGGG
+
FFFFFFFFFFF
@read2
GGGGGGGGGGG
+
FFFFFFFFFFF
@read3
GGGGGGGGGGG
+
FFFFFFFFFFF
@read4
GGGGGGGGGGGAAAAAAAAAAAAAAAAAAAAGCG
+
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```

It looks like read3 is trimmed in the middle, so we might want to reduce the allowed error rate by setting the `--error-rate` parameter to `0.05` to override the default `0.1`. It  will keep read2 and read3 from being trimmed:

```
atropos trim -a 'A{200}' --overlap 8 --no-default-adapters -se test.fq --error-rate 0.05

@read1
GGGGGGGGGGG
+
FFFFFFFFFFF
@read2
GGGGGGGGGGGAAAAAAAAGAAAAAAACAAAAA
+
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
@read3
GGGGGGGGGGGAAAAAAAAAAAAAAAAAAAAGC
+
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
@read4
GGGGGGGGGGGAAAAAAAAAAAAAAAAAAAAGCG
+
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```

That's optional, but might worth benchmarking to determine the optimal error rate.

vladsaveliev:
After looking more, learned that another atropos option `--nextseq-trim` does polyG trimming (with zero error tolerance), which probably explains why the command worked in your tests.

chapmanb:
Vlad -- wow, thanks for this forensic work and testing. I'm still confused about the behavior of the `$` anchor, but from your examples we are doing the right thing and avoiding trimming in the middle of the reads which is what we wanted to avoid. I'm fine with read3 being trimmed within the error parameters and mostly wanted to avoid trimming polyX regions internally to sequences that have non-polyX 3' anchors.

This is meant to supplement nextseq-trim with some error flexibility and also deal with other polyX issues which lead to slow runtimes with poor quality 3' noise that piles up in genomic polyX regions.

Thank you again for all this testing and the fix.

vladsaveliev:
Thanks for merging Brad!

> baffled by the behavior of `$` 

Think of $ as a regex end of line. `A{5}$` translates into "`AAAAA` followed by the end of line", thus it won't work unless you have _exactly_ 5 symbols in the end of read:
`TAAAA` -> `atropos -a "A{4}$"` -> `T`
`TAAAA` -> `atropos -a "A{5}$"` -> `TAAAA`

> and why it doesn't trim internally without it

It does trim in the middle, but you required the adapter to be 200 bases long! And it just never happens in the middle. However, when it's in the end, it allows partial overlaps. 

It all makes total sense if you think about that. Actially it's easier to think in terms of real adapters rather than polyX stretches. Say you want to trim ADAPTER sequence from all reads. You apply `-a "ADAPTER"` and expect it to do the following:
```
GGGADAPTERGGG -> GGGGGG
GGGADAPGGG -> GGGADAPGGG
GGGGGGADAP -> GGGGGG
```
And that's exactly what it does!

chapmanb:
Vlad;
Thanks so much for the explanation. That makes good sense now. It's totally coincidence that my approach of having a huge number of potential polyX's (to capture larger stretches) also anchors to the 3' end but cool that it does. Thanks again for putting this straight and helping improve the trimming.

PR #2434, opened 2018-07-07
python3 transition - print lacked ()s

smoe:
Cheers,
Steffen

### DIFF
--- a/scripts/bcbio_nextgen.py
+++ b/scripts/bcbio_nextgen.py
@@ -26,13 +26,14 @@ Usage:
      -s scheduler for ipython parallelization (lsf, sge, slurm, torque, pbspro)
      -q queue to submit jobs for ipython parallelization
 """
+from __future__ import print_function
 import os
 import argparse
 import sys
 
 from bcbio.setpath import prepend_bcbiopath
 prepend_bcbiopath()
-
+  
 from bcbio import install, utils, workflow
 from bcbio.illumina import machine
 from bcbio.distributed import runfn, clargs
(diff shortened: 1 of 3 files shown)


smoe:
Sigh. Maybe better wait for these Python3 bits until you are all confident that you want to make that transition.

chapmanb:
Steffen -- apologies for the regression and thanks for keeping this up to date. We definitely want to move to python3 so having this in place will really help. I appreciate you being on top of this.

smoe:
Uh, sorry, my "sigh" was about the technical weirdness of the Python transition, not about the Python3 incompatibility in your script. To be a bit more specific: I had just found out that apparent syntactical "ok"s have different semantics. The `import from __future__` fixes that, just please be careful. For everyone to suffer along:
```
Python 2.7.15 (default, May  1 2018, 05:55:50)
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> print "a","a" # not ok in Python 3
a a
>>> print ("a","a") # ok in Python 3 but with different semantic in Python 2
('a', 'a')
>>> from __future__ import print_function
>>> print ("a","a") # makes Python 2 behave like Python 3
a a
>>>

Python 3.6.6 (default, Jun 27 2018, 14:44:17)
[GCC 8.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from __future__ import print_function
>>> print ("a","a") # the import from __future__ is also ok in Python 3
a a
```
The ">>" redirection apparently needs some extra lines to reimplement in a cross-version-compatible way. Coming.

Best,

Steffen

PR #2437, opened 2018-07-09
Eliminated "print >>" for python3

smoe:
I should have rebased the changes over here and not work with the web interface. Hope you don't mind too much.


chapmanb:
Steffen -- thanks again for being on top of all these. Much appreciated.

PR #2470, opened 2018-07-27
TitanCNA: use correct parameters for WES runs

lbeltrame:
According to the documentation
(https://github.com/gavinha/TitanCNA/tree/master/scripts/R_scripts) the
model needs different parameters depending on variance in read data and
depending on the data type.

This commit adds the two recommended parameters for non-WGS runs.


chapmanb:
Luca -- thanks much for catching this. I pushed a follow up to adjust slightly but with the same functionality. Thanks again for all this testing.

PR #2481, opened 2018-08-09
updated dbNSFP vcfanno config for v3.5a

matthdsm:
Updated config for dbNSFP v3.5a


matthdsm:
relates to https://github.com/chapmanb/cloudbiolinux/pull/271

matthdsm:
I'm not sure if the version in `hg38/GRCh37_resources.yaml` should be bumped, since the filename doensn't change, so I left that alone for now.

Cheers
M

chapmanb:
Matthias -- thanks much for keeping this up to date. You're exactly right that we don't need to bump anything in the resources.yaml since file references did not change. Thanks again.

PR #2482, opened 2018-08-10
bumped genome resources voor VEP v93

matthdsm:


chapmanb:
Matthias -- thank you for keeping these all up to date and synchronized.

PR #2483, opened 2018-08-11
Set qualimap multiqc coverage cutoffs based on sample depths

vladsaveliev:
As discussed, automatically setting thresholds when qualimap is in enabled tools.

I also had to move the `strucrual.regions` import out of the variation.coverage module to avoid circular imports `variation/coverage.py > structural/__init__.py -> structural/cn_mops.py -> structural/shared.py bam/callable.py -> variation/coverage.py`

Run a test, seems to be working.

### DIFF
--- a/bcbio/bam/callable.py
+++ b/bcbio/bam/callable.py
@@ -25,6 +25,7 @@ from bcbio.pipeline import shared
 from bcbio.pipeline import datadict as dd
 from bcbio.variation import coverage
 from bcbio.variation import multi as vmulti
+from bcbio.structural import regions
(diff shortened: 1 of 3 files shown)


vladsaveliev:
Hmm, the Travis fail doesn't seem to be me?

chapmanb:
Vlad -- this is awesome, thanks so much for tackling it. Apologies about the Travis issues, you're right that it didn't fail due to your PR. I've been battling it trying to get runs in under the Travis timelimits. Thank you again for porting this over.

PR #2485, opened 2018-08-16
updated genome resources for dbSNP, clinvar

matthdsm:
Updates for clinvar/dbSNP
This makes sure the versions in the VEP cache and the resources are more aligned

ref:https://github.com/chapmanb/cloudbiolinux/pull/273
Cheers
M


chapmanb:
Thanks again for keeping these in sync. Much appreciated.

PR #2497, opened 2018-08-30
dbNSFP: Update hg19_pos col name

matthdsm:
The dash `-` in the column name causes vcf2db to exclude the data from the db for some reason. Removing it should fix the issue.

Cheers
M


chapmanb:
Matthias -- thanks much for being on top of this and for the fix.

PR #2499, opened 2018-09-03
Viral check: do for normals

vladsaveliev:
Often we want to see if the virus is present in a tumor sample only (which would supports its oncogenicity). How about this quick fix to check and report the viral content in normals too?

### DIFF
--- a/bcbio/pipeline/qcsummary.py
+++ b/bcbio/pipeline/qcsummary.py
@@ -107,7 +107,7 @@ def get_qc_tools(data):
         to_run += ["qsignature", "variants"]
         if vcfanno.is_human(data):
             to_run += ["contamination", "peddy"]
-        if vcfutils.get_paired([data]):
+        if vcfutils.get_paired_phenotype(data):
             to_run += ["viral"]
         if damage.should_filter([data]):
             to_run += ["damage"]


chapmanb:
Vlad -- thanks for this, happy to have more helpful QC for viral checks and include the normals as well. Much appreciated.

audyavar:
How do we interpret the columns of *-gdc-viral-completeness.txt? What do the columns depth, 1x, 5x, 25x denote?

chapmanb:
Thanks for the questions. Those are the outputs of mosdepth coverage calculations and you can read more about the output files here:

https://github.com/brentp/mosdepth#thresholds

Hope this helps

vladsaveliev:
@audyavar, the values under 1x, 5x, 25x columns represent the percentage of the viral sequence covered by reads at least 1x depth, 5x depth, 25x depth. 

The idea is that a real viral contamination will have a certain level of coverage across a large part of the viral genome. In contrast with a false positive which can have a high read support, but concentrated in a very small region, which just happened to be similar to some human sequence.
  
Hope this is helpful!

audyavar:
Thanks Brad. That is very helpful! Can you comment on what threshold and values do you consider to be viral positive at the various coverage cut-offs?

Sincerely,

Akshata

PR #2502, opened 2018-09-04
Fix parsing rRNA stats

vladsaveliev:
rRNA stats don't show up in the MultiQC report because of broken parsing. That's how they end up in `*_bcbio.txt`:

screen shot 2018-09-04 at 17 23 06

This fixes it by stripping `\n` in the end of lines.

cc @ohofmann


chapmanb:
Vlad and Oliver -- thanks much for catching this and for the fix.

PR #2533, opened 2018-10-05
Fix getting input files for non-flat input structure

bogdang989:
Hi @chapmanb I was testing the test-bcbio-cwl workflows and some of them fail on rabix and SBG platform on multiqc step. I think the reason for this is because the secondary input files from some qc steps are created in subfolders relative to the primary file. When these files are mounted in the docker container in cwltool the structure is flattened, while in rabix/sbg docker containers the original structure is preserved. This modification should handle both cases. Tested with rabix and cwltool.


chapmanb:
Bogdan;
Thanks much for the fix for nested cases. Apologies, I hadn't seen that previously and appreciate you generalizing to work on SBG as well.

PR #2552, opened 2018-10-28
Update bcbio/cwl/hpc.py

apastore:
add lsf default_config and fix job-id-regex


chapmanb:
Thanks so much for these fixes; I appreciate you adding and testing this.

Issue #2570, opened 2018-11-07
adding GnomAD to gemini config


matthdsm:
Hi Brad,

how would you feel about supplanting ExAC with Gnomad Exome in the default gemini config? This would remove the need for some `postannotation` scripting, and the users would get updated frequencies.
Alternatively, instead of removing ExAC, we could just add GnomAD?

Cheers
M

chapmanb:
Matthias and Luca;
Thanks for starting this conversation. I'm definitely open to replacing ExAC with GnomAD. My main concern is around file sizes and how to manage hg38 support. Right now I believe it's still build 37 only (http://gnomad.broadinstitute.org/downloads) and I didn't see any hg38 liftover resources prepared for release 2.1 with a quick search. For sizes, are you thinking of just the exomes (60Gb)? Annotating in genomic regions is definitely useful, but the genomes are pretty unwieldy (464Gb). What were your thoughts about what we'd provide and support? Thanks again.

lbeltrame:
Considering I'd use it mostly to evaluate variants, I'd be in support of using the exomes only.

matthdsm:
Hi Brad,

I'd suggest using the VCFs hosted by Ensembl (v2.0.1), since there's no real effort towards hg38 compatibility from Broad. As for the ExAc replacement, I'd stick to the exomes and leave the gemomes optional (as it is now). If you're okay with this, I've already got a PR ready to implement these changes.

Cheers
M

chapmanb:
Matthias and Luca;
Thanks, using the Ensembl resources makes good sense. Nice catch, I should have thought of that. Keeping up to date would be useful if Ensembl will be synchronizing soon. There are a number of calling improvements in 2.1 (https://macarthurlab.org/2018/10/17/gnomad-v2-1/). To me, the most useful are the subsets excluding cancer, case genomes, and neurology studies. This could help a lot with avoiding low frequency events from gnomAD samples correlated with different experimental conditions. Thanks so much for looking at this.

matthdsm:
I've sent an e-mail to the Ensembl helpdesk to ask if they plan to update their resources. Is it ok for now if we stick to v2.0.1? I'll submit a PR replacing ExAC with GnomAD shortly.
I this it's safe to remove the ExAC GGD recipe after this, since I don't think it'll be used anymore.

Cheers
M

matthdsm:
[...]

PR #2571, opened 2018-11-12
gnomAD exome config

matthdsm:
ref #2570 

As promised, the configs needed for gnomAD exome. I've only added stuff. I'll leave removing up to you.
I tried to keep the annotations from gnomAD as similar as possible to the existing ExAC ones, without having to use any postannotations, as all data is present in the vcf.

Cheers
M


chapmanb:
Matthias;
Thanks much for this update. I made a few changes with extra spaces in the inputs, but otherwise looks great. Do you need a separate gnomad_exome vcfanno configuration in addition to the default gemini and genome configuration? I don't mind keeping it if required, but am trying to think about having less to maintain/get out of sync over time. Thanks again.

matthdsm:
Hi Brad,

Nice catch on the whitespace, didn't notice that one. I don't really see much sense in keeping a separate exome conf file, since it's already included in the gemini conf, which can't be disabled as far as I know.
If you'd consider letting users disable the default gemini conf, while still building the db's, then I'd opt for keeping both, since the gnomAD dataset is always a huge bonus to have in the annotations.

Cheers
M

PR #2584, opened 2018-11-30
Fix qsignature for multiqc

vladsaveliev:
Fixing uploading qsignature .ma file to propagate it into multiqc_bcbio:
- summarize qsignature _before_ running MultiQC so all files are in place
- add qsignature.ma into multiqc file list
- multiqc file_list_final: support dirs to propagate mixup_check with qsignature.ma file

Though later realized that qsignature reporting in MultiQC is commented out: https://github.com/MultiQC/MultiQC_bcbio/pull/24


chapmanb:
Thanks for fixing this, I'm not sure about the current multiqc status but will help figure it out. Thanks for working on it.

PR #2599, opened 2018-12-14
WIP Avoid adding directories to secondaryFiles with class File

bogdang989:
Depending on the file structure for inputs, directories might be present in the same folder with primary file (which actually happens in SBG execution) but they are added with with `"class": "File"`. This fix avoids adding these directories to the secondary file list.


chapmanb:
Thanks so much for testing on SBG and for the fix. Much appreciated.

PR #2604, opened 2018-12-19
Extend check for directory in secondaryFiles

bogdang989:
Update for #2599
Extend the check to all cases where secondaryFiles are populated.


chapmanb:
Thank you for catching all these additional cases. I'm excited to have this working cleanly on SBG.

Issue #2616, opened 2019-01-10
Null values instead of numeric values in multiqc_data_final.json


roland-ewald:
[A recent change to `qc/multiqc.py`'s `_work_path_to_rel_final_path`](https://github.com/limbus-medtec/bcbio-nextgen/commit/ed8ccdb6a9a49524076a210ff7b165623a32d436#diff-7e97be75fbfe2fe0ba2f56111d713646R190) makes it return `None` whenever `path` is not a string, which means that [`utils.walk_data` invocations walking over the `multiqc` JSON data](https://github.com/bcbio/bcbio-nextgen/blob/master/bcbio/qc/multiqc.py#L148) will replace all numeric values by `None`.

chapmanb:
Fixed in #2617

PR #2617, opened 2019-01-10
#2616: fix multiqc_data_final.json null values

roland-ewald:
`_work_path_to_rel_final_path` should return `path` instead of `None` whenever
path is not a string, as `utils.walk_data` invocations walking over
the `multiqc` JSON data will otherwise replace all non-string (e.g.
numeric) values by `None`.


chapmanb:
Thank you also for this investigation and fix. Please let us know if you hit any other issues at all.

roland-ewald:
Will do :-) Thanks for merging!

Issue #2618, opened 2019-01-10
Output qc/variants/*_bcftools_stats.txt not created anymore


roland-ewald:
These files are not created anymore because `get_active_vcinfo` in [`qc/variant.py`](https://github.com/bcbio/bcbio-nextgen/blob/master/bcbio/qc/variant.py#L115)
is called with the default value for `use_ensemble` (`True`), and with a [recent change](https://github.com/bcbio/bcbio-nextgen/commit/1ee74d3b38cc178a2e33b323b192fa7bc8e492e3#diff-90458b3fe804d92ceebabda7cec07cb4) this will 
cause the method to return an empty array instead of the first element on invocation
when running a non-ensemble setup (see below for details).

It seems there are two ways to fix this: either removing the default value for the `use_ensemble` 
argument and then using the actual value for all invocations of `get_active_vcinfo`, 
or adjusting `get_active_vcinfo` to fail gracefully in this specific case. 

I'm not sure what the best way is to fix this. The latter option seemed to be the easier solution 
(also in the amount of required code changes), so I chose that one.

### How to reproduce

Running bcbio with a `SAMPLE.bam` file via

````bash
bcbio_nextgen.py --only-metadata -w template test.bcbio.yaml test.bcbio.csv input/SAMPLE.bam
mkdir -p work && cd work
bcbio_nextgen.py -n 2 ../test_bcbio/config/test_bcbio.yaml
````

with the files as outlined below.

It works, but the file `work/final/SAMPLE/qc/variants/SAMPLE_bcftools_stats.txt` will be missing from the results.

#### `test.bcbio.csv`

````csv
samplename,description,batch,phenotype,sex,variant_regions,coverage
SAMPLE,SAMPLE,,,,config/test.bed,config/test.bed
````

#### `test.bcbio.yaml`

````yaml
---
details:
  - analysis: variant2
    genome_build: GRCh37
    algorithm:
      coverage_interval: regional
      mark_duplicates: true
      variantcaller: [gatk-haplotype]
      tools_on: []
      tools_off: [bwa-mem, gemini, fastqc, coverage_qc]
      # ...
upload:
  method: filesystem
  dir: final
resources:
# ...
````

chapmanb:
Fixed in #2619

PR #2619, opened 2019-01-10
#2618: fix qc/variants/*_bcftools file creation

roland-ewald:
These files are not created anymore because `get_active_vcinfo`
returns an empty array instead of the first element on invocation
when running a non-ensemble setup.

Instead, it now fails 'gracefully' by attempting to find an entry for
the 'ensemble' variant caller, but if it does not find one it will still
return the first non-ensemble one.

### DIFF
--- a/bcbio/qc/variant.py
+++ b/bcbio/qc/variant.py
@@ -117,9 +117,10 @@ def get_active_vcinfo(data, use_ensemble=True):
     """
     active_vs = _get_variants(data)
     if len(active_vs) > 0:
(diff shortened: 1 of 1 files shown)


chapmanb:
Thanks so much for this report and this fix; this approach is perfect. I appreciate your help diagnosing and fixing the issue.

roland-ewald:
Hi Brad, thank you for merging this so quickly, and for the kind words. 
It's a bliss to contribute to a project that is so well maintained!

PR #2644, opened 2019-01-23
Fix HISAT2 firststrand strandedness for single-end samples

mjsteinbaugh:
`hisat2.py`: The if/else statement to handle the `strandedness` argument for stranded single-end samples had a typo that results in the pipeline always running as `secondstrand`, even when the user requests `firststrand`.


chapmanb:
Mike;
Thanks so much for catching this. I tried to make the whole thing a bit more resilient here since that's an easy typo to make, so hopefully will all work cleanly for multiple cases now. Thanks again.

PR #2648, opened 2019-01-25
move VEP cache download from ftp > http

matthdsm:
FTP is blocked by a lot HPC's due to some safety concerns, changing the URL to HTTP fixes the download issues.

Cheers
M


chapmanb:
Matthias -- thanks much for the update. It's great to have this working in more environments.

PR #2677, opened 2019-02-09
BPI clean up: 1. set tmp dir for JVM 2. avoid redundant second run

vladsaveliev:
1. Setting up the JVM temporary directory explicitly to avoid "No space left on device" error. Thanks @ohofmann 
2. For tumor/normal pairs, BPI was run twice on the same VCF because it was called in the loop over all samples. Add check that sample is tumor to assure it runs once.
3. Write stdout into a file because it tend to become very verbose for large datasets.


vladsaveliev:
Hi Rory! Thanks for reviewing. Agree - added (though that file is just a useless stdout log)

roryk:
Awesome, thanks so much. I think I fixed the bug that was causing travis to fail, I'll merge this in when it finishes up.

PR #2681, opened 2019-02-14
barcodes_samples now accept files like:

lpantano:
Just doing this merge through pull request because it was quite big.

barcodes_samples now accept files like:

AAAAAAA,sample1

That will be used to name the cells from that sample in the final
tagcounts.mtx.metadata


roryk:
Thank you for this! This is a useful feature.

lpantano:
I hope so, the singlecellexperiments is not on yet, but working on that.

PR #2685, opened 2019-02-19
VarDict: handle "-f" (allele freq) set in resources>vardict>options

vladsaveliev:
"-f" when set in `resources/vardict/options` would clash with the hardcoded "-f {freq}" defined based on `algorithm/min_allele_fraction`. 

Adding a check to avoid duplicating the option and allow overriding min_allele_fraction for VarDict.


chapmanb:
Vlad -- thanks so much for the improvement and the clear documentation about how the setting gets applied.

PR #2732, opened 2019-03-25
install.py: fix typo

### LINKED ISSUE #2731: Should upgrade require installing libgcc-ng unconditionally?
Given how awful the dependency solving times of conda are (I guess I got spoiled from libsolv and similar projects) I wonder if first checking if a newer version is available is preferable and only then perform the install.

On this (admittedly old) bcbio install the first two conda runs, without even arriving to the bcbio upgrade step, take more than an hour to run.

lbeltrame:
Fixes #2731.


chapmanb:
Luca -- thanks so much for catching this and for the quick fix.

Issue #2762, opened 2019-04-08
Handle mirdeep crash for case when there are many potential precursors


rdocking:
At this issue: 

https://github.com/lpantano/mirdeep2_core/issues/3

I figured out the underlying cause of why mirdeep2 was halting prematurely when run with large miRNA datasets. The fix is described at the linked issue - briefly, there is a hard-coded limit to the number of miRNA precursors that needs to be over-ridden to allow for processing large data sets. I'll submit a pull request that addresses the issue.

lpantano:
thanks!

PR #2763, opened 2019-04-08
Add -g -1 option to prevent early exit in mirdeep2

rdocking:
Addresses https://github.com/bcbio/bcbio-nextgen/issues/2762


lpantano:
thanks!

PR #2766, opened 2019-04-09
Vep 96

matthdsm:
Hi Brad,

I've made some updates following the release of vep 96.

Cheers
M


chapmanb:
Matthias;
Thanks again for the updates and fixes for synchronize with the latest VEP.

PR #2773, opened 2019-04-11
Eliminated warning for Sphinx Doc generation

smoe:
Sorry for that piece of irrelevance.

You referenced CloudBioLinux. I would (obviously) very much like like to bring the Debian package to a stage that it would qualify to be accepted here. At what stage would you consider adding such a reference?

Best,
Steffen


roryk:
Thank you for this! The documentation does not get as much love as we would like.

PR #2774, opened 2019-04-11
Fixing indication of regular expressions

smoe:
Avoids deprecation warnings. When running python with -W all, like when running autotests of the bcbio Debian package, one can otherwise observe the following:
```
bcbio/broad/__init__.py:246
  /bcbio/.pybuild/cpython3_3.7_bcbio/build/bcbio/broad/__init__.py:246: DeprecationWarning: invalid escape sequence \d
    pat = re.compile('([\d|\.]*)(\(\d*\)$)')  # matches '1.96(1510)'
bcbio/structural/cn_mops.py:75
  /bcbio/.pybuild/cpython3_3.7_bcbio/build/bcbio/structural/cn_mops.py:75: DeprecationWarning: invalid escape sequence \w
    return re.sub("[^\w.]", '.', name)
bcbio/structural/delly.py:82
  /bcbio/.pybuild/cpython3_3.7_bcbio/build/bcbio/structural/delly.py:82: DeprecationWarning: invalid escape sequence \.
    cmd = ("bcftools view {bcf_file} | sed 's/\.,\.,\././' | bgzip -c > {tx_out_file}")
```

### DIFF
--- a/bcbio/broad/__init__.py
+++ b/bcbio/broad/__init__.py
@@ -240,7 +240,7 @@ class BroadRunner:
         cl += ["--version"]
         p = subprocess.Popen(cl, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
         # fix for issue #494
-        pat = re.compile('([\d|\.]*)(\(\d*\)$)')  # matches '1.96(1510)'
+        pat = re.compile(r'([\d|\.]*)(\(\d*\)$)')  # matches '1.96(1510)'
         m = pat.search(p.stdout.read())
         version = m.group(1)
         self._picard_version = version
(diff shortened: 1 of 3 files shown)


roryk:
Thanks for these-- I had no idea python had a `-W all` option, that is nice. Does the delly one actually apply?

smoe:
I think it does, even if the Python compiler may have found it by accident. It has a very long string representing the command line with a serial piping. You intend to pass to sed the backslash followed by a period, not the non-existing control character `\.`. I just could not use the "r" prefix to the string as I did for the other two cases but fell back to the traditional way to express a backslash in a string.

smoe:
The page https://docs.python.org/3.7/library/warnings.html?highlight=warning#module-warnings describes how to specify filters for warnings.

roryk:
Thank you for this one too-- I fixed these. Again, sorry for not merging this in, I didn't want to make it a hassle to dump/test the other more escaping ones.

PR #2776, opened 2019-04-11
Doc fixes

smoe:
A few subtle differences.


roryk:
Thank you, having these fixed is so great.

PR #2777, opened 2019-04-11
Series of spelling errors in documentation.

smoe:
@mr-c  once upon a time fixed these for the Debian package. And then these have apparently
been forgotten about.  That is the problem with 24h days.
There are ways that are past my routine git-with-github usage to adjust the author of the patch. I have cross-checked with already-bcbio-contributor Michael. He would be happy to have this patch under his name but not ultimately desparate if this arrives with my ID.


roryk:
Thank you for these!

PR #2781, opened 2019-04-15
Fix bug in extracting rRNA features from gtf

vladsaveliev:
Fixes the bug where instead of a geneid/txid pair, there was double geneid output, leading to always 0.0 rRNA rate stat.

cc @ohofmann


roryk:
Hi Vlad, I am a moron. Thank you so much for finding this and fixing it-- I haven't run a bulk RNA-seq in a while on a genome that has rRNA annotated, so I missed this.

PR #2786, opened 2019-04-16
Fix urllib usage in install.py

lbeltrame:
`getheader()` is no more in Python 3, use `get()` instead.


roryk:
Thanks Luca! It is so helpful to have folks fixing these issues before they arise. We really appreciate it.

lbeltrame:
This doesn't look like it was merged. Was this intentional?

roryk:
Oops, sorry

roryk:
Thanks! Sorry, I must have clicked the wrong button.

PR #2790, opened 2019-04-17
run_test.sh - allow for py.test-3 binary, beautification

smoe:
After updating master, most of my changes for Python V3 were already done :o) I found another regexp missing the "r" prefix at least.

PYTEST can now be specified externally and it looks for the distribution's default py.test install paths. I can understand if you frown about the iteration of py.test locations. It would then be nice if you could still accept the setting via an environment variable.

Cheers,

Steffen

### DIFF
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -53,15 +53,10 @@ if [ -d "../bcbio/pipeline" ]; then
 fi
 
 if [ -z "$PYTEST" ]; then
-	for p in "$BCBIO_DIR/py.test" /usr/bin/py.test-3 /usr/bin/py.test
-	do
-		if [ -x "$p" ]; then
-			PYTEST="$p"
-			break
-		fi
-	done
-	if [ -z "$PYTEST" ]; then
-		echo "E: Could not identify py.test to execute."
-	fi
+	PYTEST="$BCBIO_DIR/py.test"
+fi
+if [ ! -x "$PYTEST" ]; then
+	echo "E: py.test (at '$PYTEST') not found or not executable."
+	exit 1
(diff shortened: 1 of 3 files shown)


tests/run_tests.sh
```diff
@@ -17,17 +9,59 @@ if [ -n "$1" ]; then
...
 export PYTHONNOUSERSITE=1
+
 # Ensure version.py exists in raw cloned bcbio directory
 if [ -d "../bcbio/pipeline" ]; then
 	[ -f ../bcbio/pipeline/version.py ] || touch ../bcbio/pipeline/version.py
 fi
-"$BCBIO_DIR/py.test" -p no:cacheprovider -p no:stepwise -v -s -m ${MARK} "$@"
+
+if [ -z "$PYTEST" ]; then
+	for p in "$BCBIO_DIR/py.test" /usr/bin/py.test-3 /usr/bin/py.test
```
roryk: Thanks! It is probably better to just check the `py.test` in bcbio and then bail out if it isn't there; falling back to system installed pythons has been a source of pain in the past so we try to keep everything in bcbio completely independent from whatever is installed on the system.

smoe: I think to have simplified it accordingly. The PYPATH environment is still allowed to contribute, was not sure if you mind that. You know about the background of this patch to have it all compatible for Debian - and for it there is no difference between bcbio's python and the system-installed one - except for the path.

roryk:
Thanks! I think we're all set with the python3 support and I fixed the missing `r` you found. Thanks again for all of these improvements, sorry for not merging these in and really sorry for taking forever.

PR #2791, opened 2019-04-17
Added series of "r" prefixes to REs

smoe:
Please have a second look at the simplified invocation of sed for bcbio/rnaseq/bcbiornaseq.py, sparing the use of backslashes. I have not tested this locally, but digging deep in my past this should work.


roryk:
Thank you, I fixed all of the `re` ones. I wasn't sure if the ones replacing sed commands would work, so I left them out. Sorry for not merging this in, but thank you very much for fixing these.

Issue #2806, opened 2019-04-30
Stream closed error with bcbio ApplyBQSRSpark 


Fazulur:
Dear Bcbio Team,

I upgraded bcbio recently to v1.1.5 and I am getting below error when running gatk-variant pipeline.

[April 28, 2019 10:11:25 AM AST] org.broadinstitute.hellbender.tools.spark.ApplyBQSRSpark done. Elapsed time: 0.15 minutes.
Runtime.totalMemory()=874512384
**htsjdk.samtools.util.RuntimeIOException: java.io.IOException: Stream closed**
        at htsjdk.samtools.IndexStreamBuffer.readFully(IndexStreamBuffer.java:23)
        at htsjdk.samtools.IndexStreamBuffer.readLong(IndexStreamBuffer.java:62)
        at htsjdk.samtools.AbstractBAMFileIndex.readLong(AbstractBAMFileIndex.java:436)
        at htsjdk.samtools.AbstractBAMFileIndex.query(AbstractBAMFileIndex.java:311)
        at htsjdk.samtools.CachingBAMFileIndex.getQueryResults(CachingBAMFileIndex.java:159)
        at htsjdk.samtools.BAMIndexMerger.processIndex(BAMIndexMerger.java:43)
        at htsjdk.samtools.BAMIndexMerger.processIndex(BAMIndexMerger.java:16)
        at org.disq_bio.disq.impl.file.IndexFileMerger.mergeParts(IndexFileMerger.java:90)
        at org.disq_bio.disq.impl.formats.bam.BamSink.save(BamSink.java:132)
        at org.disq_bio.disq.HtsjdkReadsRddStorage.write(HtsjdkReadsRddStorage.java:225)
        at org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSink.writeReads(ReadsSparkSink.java:155)
        at org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSink.writeReads(ReadsSparkSink.java:120)
        at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.writeReads(GATKSparkTool.java:361)
        at org.broadinstitute.hellbender.engine.spark.GATKSparkTool.writeReads(GATKSparkTool.java:349)
        at org.broadinstitute.hellbender.tools.spark.ApplyBQSRSpark.runTool(ApplyBQSRSpark.java:90)
[...]

PR #2810, opened 2019-05-05
Switch to using packaging.version.

roryk:
LooseVersion is apparently not intended for outside code use,
and there are some issues with comparing versions with and
without strings in LooseVersion that they will not fix.

See https://bugs.python.org/issue14894 for the discussion.

packaging is a third-party package that does version checking, it
is python specific and follows this: https://www.python.org/dev/peps/pep-0440/
which looks like it should handle the cases we are looking for.

For example:

V1 = Version("4.1.0.0")
V2  = Version("4.2a")
V3 = Version("4.1.0.0a")

V2 > V1: True
V3 >= V1: False
V3 <= V1: True

This should address the issue in #2806.

@chapmanb what do you think about this way of handling it?

### DIFF
--- a/bcbio/broad/__init__.py
+++ b/bcbio/broad/__init__.py
@@ -4,7 +4,7 @@
   GATK -- Next-generation sequence processing.
 """
 from contextlib import closing
-from distutils.version import LooseVersion
+from packaging.version import Version
 import getpass
 import re
 import sys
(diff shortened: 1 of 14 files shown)


roryk:
packaging isn't installed with the base python installation, so we'd need to add it to the list of things to get installed if we go this route.

chapmanb:
Thanks Rory, this does seem like a good general way of fixing it, with the only downside being the introduction of a new dependency. I wonder if we can hack around this to avoid needing to do that work. This might be an edge case with unusual versions and we just need to figure out what GATK version gets returned in #2806 and massage it so it doesn't cause breakage?

roryk:
Closing since it looks like our usual use cases should get handled by the setuputils issue.

PR #2820, opened 2019-05-15
update dbNSFP config for v4.0

matthdsm:
Hi Brad,

I've updated the vcfanno configs to go with dbNSFP v4.0a.
All columns containing data we already source elsewhere have been omitted (eg. population db's, Ensembl, snpEff, clinvar).

Cheers
M


chapmanb:
Matthias -- thanks again for being on top of this.

Are there any opinions from other community members about the desire to move to 4.0 versus staying with the current version? There is a larger discussion about the pros and cons of this (chapmanb/cloudbiolinux#304) and we're between two minds about updated. The pros of updating are getting the latest annotations and staying current. The cons are that it's an expensive preparation process in terms of time and disk usage.

It would be great to hear what other dbNSFP users think about updating. Thanks so much.

PR #2831, opened 2019-05-23
Update vagrant_osx.sh

hackdna:
Option `--sudo` is no longer available, so writing to `/usr/local` fails. Also, installing the genomes requires 10s of GB in the VM, so adding no data by default to make installation more lightweight.


roryk:
Thanks Ilya!

PR #2839, opened 2019-05-29
Clean up downsample() and make it compatible with Python 3

hackdna:


roryk:
Thanks! Also thank you for catching that this function wouldn't work on gzipped files.

PR #2851, opened 2019-06-07
Fix description for separators flag

mjsteinbaugh:
The description recommends "semicolon" but should read "comma" instead.


roryk:
Thanks Mike!

Issue #2846, opened 2019-06-05
RFE: Mutect2: use the new strand orientation bias filter in 4.1.1.0


lbeltrame:
According to [this tutorial](https://gatkforums.broadinstitute.org/gatk/discussion/24057/how-to-call-somatic-mutations-using-gatk4-mutect2#latest) it's very useful and, according to Broad people, has not much of a performance impact and can help when handling FFPE samples.

As a downside requires one more file to track. Before I think whether to add this to bcbio, I ask, does it make sense to add this option?

chapmanb:
Luca;
Thanks for the suggestion. I agree it does make sense to look at and would be a good complement/replacement for DKFZ bias filter for detecting potential strand bias issues due to FFPEs or prep artifacts. Thanks much for looking into this.

roryk:
Thanks for the contribution Luca!

PR #2856, opened 2019-06-18
[PoC] Use Mutect2's new read model for filtering

lbeltrame:
According to the GATK's doc, Mutect2 can detect read orientation
artifacts (such as stand bias caused by FFPE, or by Illumina Novaseq)
and improve results.

This is a proof of concept (completely untested at this point). In
particular this is posted for early review.

Some important points to act on:

- Do we want this on at all times (GATK says it won't hurt performance)?
- How do we make sure this is only run for paired-end samples (see
FIXME)?
- Is the current simplistic logic for file names acceptable?

This will fix issue #2846.

### DIFF
--- a/bcbio/variation/mutect2.py
+++ b/bcbio/variation/mutect2.py
@@ -112,8 +112,8 @@ def mutect2_caller(align_bams, items, ref_file, assoc_files,
                 orientation_filter = False
 
             if gatk_type == "gatk4" and orientation_filter:
-                f1r2_file = "{}.tar.gz".format(
-                    utils.splitext_plus(align_bams[0])[0])
+                f1r2_file = "{}-f1r2.tar.gz".format(
+                    utils.splitext_plus(tx_out_file))
                 params += ["--f1r2-tar-gz", f1r2_file]
 
             # Avoid adding dbSNP/Cosmic so they do not get fed to variant filtering algorithm
--- a/bcbio/variation/mutect2.py
+++ b/bcbio/variation/mutect2.py
@@ -106,7 +106,7 @@ def mutect2_caller(align_bams, items, ref_file, assoc_files,
             params += _add_tumor_params(paired, items, gatk_type)
             params += _add_region_params(region, out_file, items, gatk_type)
 
-            if is_paired(align_bams[0]):
+            if all(is_paired(bam) for bam in align_bams):
                 orientation_filter = True
             else:
                 orientation_filter = False
--- a/bcbio/variation/mutect2.py
+++ b/bcbio/variation/mutect2.py
@@ -87,6 +87,7 @@ def mutect2_caller(align_bams, items, ref_file, assoc_files,
         paired = vcfutils.get_paired_bams(align_bams, items)
         broad_runner = broad.runner_from_config(items[0]["config"])
         gatk_type = broad_runner.gatk_type()
+        f1r2_file = None
         _prep_inputs(align_bams, ref_file, items)
         with file_transaction(items[0], out_file) as tx_out_file:
             params = ["-T", "Mutect2" if gatk_type == "gatk4" else "MuTect2",
@@ -103,6 +104,13 @@ def mutect2_caller(align_bams, items, ref_file, assoc_files,
                 params += ["--read-validation-stringency", "LENIENT"]
             params += _add_tumor_params(paired, items, gatk_type)
             params += _add_region_params(region, out_file, items, gatk_type)
+
+            #FIXME: How to detect paired vs single end?
+            if gatk_type == "gatk4":
+                f1r2_file = "{}.tar.gz".format(
+                    utils.splitext_plus(align_bams[0])[0])
+                params += ["--f1r2-tar-gz", f1r2_file]
+
             # Avoid adding dbSNP/Cosmic so they do not get fed to variant filtering algorithm
             # Not yet clear how this helps or hurts in a general case.
             #params += _add_assoc_params(assoc_files)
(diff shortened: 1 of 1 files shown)


chapmanb:
Luca;
Thanks much for working on this, I appreciate you digging into it. My take on your questions:

- I'd have it on all the time. If we find performance issues then can have a tools_on/tools_off based approach.
- For paired end samples, you can check for `paired.normal_bam` and skip running it if it's not present. Other tools do this checking as well, for instance strelka2 (https://github.com/bcbio/bcbio-nextgen/blob/7fb884df990610c553043d5ed9cdf2b4b35f178a/bcbio/variation/strelka2.py#L29).
- All of the file naming looks great to me.

Thanks again for implementing this.

lbeltrame:
Sorry, by "paired" I meant "paired end [reads]", since it uses the F and R read orientation to do its job. If we assume everything is paired-end, I can just remove the comment.

I'll try to test this later this week.

chapmanb:
Luca;
Sorry about the non-helpful comments about the wrong paired check. We have a function in `bcbio.bam` to check for paired versus single end from the BAM files:

https://github.com/bcbio/bcbio-nextgen/blob/0dd6ab7002d6fe3c66e7c59dc19011b017646c19/bcbio/bam/__init__.py#L43

Hopefully that works for you. Thanks again for working on this.

scatreux:
Hi Luca
We'd be interested to test the GATK OB filter from BCBIO. When do you think you can have a version ready for us to run?
thanks!
Severine

lbeltrame:
Whenever I get around to see if it is actually running ;) I need to implement what Brad suggested first. 
@chapmanb Is there any *super small* test dataset among those used by bcbio to test this stuff?

scatreux:
Thanks Luca

Sounds like it may be ready pretty soon, maybe a week or 2?

scatreux:
BTW, we have noted the following sequence of 3 steps for the OB filter:
First, run Mutect2 with the --f1r2-tar-gz argument. 
Next, pass this raw data to LearnReadOrientationModel:
Finally, pass the learned read orientation model to FilterMutectCallswith the -ob-priors argument.

Are you following this sequence in BCBIO?

lbeltrame:
The code might be a little confusing, but that's exactly what it's doing (or supposed to do) when MuTect2 is run.

chapmanb:
Luca and Severine;
Thanks again for looking into this. For running basic tests, `./run_tests.sh cancermulti` runs a small tumor/normal pair through many callers and you can modify the input configuration in `./tests/data/automated/run_info-cancer.yaml` to only have mutect2 for testing purposes. Thanks again.

lbeltrame:
Does that test need bcbio installed? I don't (obviously) have that on my dev machine.

chapmanb:
Luca -- it needs to have the bcbio code and tools installed, but not any data. The test repo has it's own small genome and associated data so you don't need a full install. This is the framework we use for testing on travis. Hope this helps.

lbeltrame:
My first run didn't quite end up like I expected. Most MuTect2 calls fail with

```
java.lang.IllegalArgumentException: errorRate must be good probability but got NaN
        at org.broadinstitute.hellbender.utils.Utils.validateArg(Utils.java:730)
        at org.broadinstitute.hellbender.utils.QualityUtils.errorProbToQual(QualityUtils.java:225)
        at org.broadinstitute.hellbender.utils.QualityUtils.errorProbToQual(QualityUtils.java:209)
        at org.broadinstitute.hellbender.tools.walkers.mutect.filtering.Mutect2FilteringEngine.lambda$applyFiltersAndAccumulateOutputStats$13(Mutect2FilteringEngine.java:176)
        at java.util.Optional.ifPresent(Optional.java:159)
        at org.broadinstitute.hellbender.tools.walkers.mutect.filtering.Mutect2FilteringEngine.applyFiltersAndAccumulateOutputStats(Mutect2FilteringEngine.java:174)
        at org.broadinstitute.hellbender.tools.walkers.mutect.filtering.FilterMutectCalls.nthPassApply(FilterMutectCalls.java:142)
        at org.broadinstitute.hellbender.engine.MultiplePassVariantWalker.lambda$traverse$0(MultiplePassVariantWalker.java:40)
        at org.broadinstitute.hellbender.engine.MultiplePassVariantWalker.lambda$traverseVariants$1(MultiplePassVariantWalker.java:77)
        at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
        at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
        at java.util.Iterator.forEachRemaining(Iterator.java:116)
        at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
        at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:418)
        at org.broadinstitute.hellbender.engine.MultiplePassVariantWalker.traverseVariants(MultiplePassVariantWalker.java:75)
        at org.broadinstitute.hellbender.engine.MultiplePassVariantWalker.traverse(MultiplePassVariantWalker.java:40)
        at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:1039)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:139)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:191)
        at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:210)
        at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:162)
        at org.broadinstitute.hellbender.Main.mainEntry(Main.java:205)
```

However I noticed a few changes I need to make to file naming.

lbeltrame:
To be honest I'm kind of stumped as I don't understand why this is happening.

lbeltrame:
As I'm running out of time to actual work on this, I would propose to put this additional filter into a`tools_on` (like `mutect2_readmodel`) and allow people to test this better and on more diverse datasets. 

@chapmanb How does that sound? Any examples on how to check for `tools_on` anywhere?

lbeltrame:
It looks like things changed again: https://software.broadinstitute.org/gatk/documentation/article?id=24057

More stuff to implement but perhaps it will fix the issue?

lbeltrame:
Well, I think I have found the cause of the errors at least. If the context for some base combinations analyzed by `LearnReadOrientationModel` is missing (because the region is too small: my tests were with a very small targeted panel) it will create problems. 

Therefore I think that:

a. It's better to keep this as a `tools_on` option;
b. I'll try (but no guarantees) to submit a bug report to the GATK (those errors shouldn't happen)
c. I think this PR is good for review.

@chapmanb @roryk any issues you can find here, or do you think this could be merged?

lbeltrame:
Ping! Anything else that needs to be done here?

chapmanb:
Luca -- thanks so much for adding this and for making it optional with `tools_on` while testing. I'd meant to have a deeper look into it to see if I could avoid or work around the small region issues but also don't want to delay others testing it so happy to get it in so others can work with it. Thank you again for working on this.

PR #2879, opened 2019-07-15
decouples VEP installation from bcbio run

naumenko-sa:


roryk:
Thanks Sergey! Build finally passed, conda has some non-deterministic high-memory behavior that we kept running into.

PR #2882, opened 2019-07-17
described how to load scRNA-seq results from bcbio into Seurat

naumenko-sa:


roryk:
Thanks Sergey, sorry for being slow about merging this in. Another way to do it is to just read in the matrix with readMM and set the column and rownames in that object. Making it follow the 10x naming conventions works too though.

Issue #2402, opened 2018-05-22
Support options for CNVkit segmentation


lbeltrame:
Use case: using other segmentation methods which may perform better on smaller targeted panels, for example.
I think the usual approach like done for other tools would fit, that is having to add the relevant bits to the YAML configuration.

chapmanb:
Luca;
Thanks so much for the suggestion, the latest development version of bcbio now supports:
```
resources:
  options:
    cnvkit_segment: [--method, haar]
```
This is undocumented as I'd rather not have this used in people's production environments but agree it can be useful for exploring and tweaking. If you find options that work better for different input types, I'd love to incorporate test cases and make these different segmentation approaches happen automatically within bcbio logic. Thanks again for exploring this.

PR #2894, opened 2019-07-30
Fix setting CNVkit options in configuration

lbeltrame:
In #2402, an (undocumented by design) option was set to allow the use of
alternative segmentation methods by CNVkit. However, the logic of
extracting configuration was swapped, thus options were never set.

For a typical YAML:

```yaml
resources:
  options:
    cnvkit_segment: [--drop-low-coverage, -m, hmm]
```

Before:

```python
resources = config_utils.get_resources("cnvkit_segment", data["config"])
user_options = resources.get("options", [])
```

The first option would not pick up "cnvkit_segment", but only the
fallback inside `get_resources`, so `user_options` would always be
empty.

After:

```python
resources = config_utils.get_resources("options", data["config"])
user_options = resources.get("cnvkit_segment", [])
```

And like this they are correctly set.


roryk:
Hi Luca,

Thanks so much for finding and fixing this. It is incredibly helpful having folks contribute back like this, we really, really, really appreciate it. Thank you!

PR #2937, opened 2019-09-11
Add support for germline resources (from gnomAD) in Mutect2

lbeltrame:
If available, the gnomAD exome file will be passed to Mutect2 to avoid
calling sites which are already known to be germline. This will
be mostly useful for tumor-only runs because known sites are not called
at all, but could also be useful for paired runs.

### DIFF
--- a/bcbio/pipeline/datadict.py
+++ b/bcbio/pipeline/datadict.py
@@ -207,6 +207,9 @@ LOOKUPS = {
     "tools_off": {"keys": ["config", "algorithm", "tools_off"], "default": [], "always_list": True},
     "tools_on": {"keys": ["config", "algorithm", "tools_on"], "default": [], "always_list": True},
     "cwl_reporting": {"keys": ["config", "algorithm", "cwl_reporting"]},
+    "gnomad_exome_file": {"keys": ["genome_resources", "variation",
+                                   "gnomad_exome"],
+                          "checker": file_exists}
 }
 
 def get_background_cnv_reference(data, caller):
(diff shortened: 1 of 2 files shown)


roryk:
Thanks for putting this together Luca! It is so awesome having folks contribute back ideas and code. I am testing this out today on the somatic GIAB mix and will merge it in after.

roryk:
Hi Luca,

Thanks for this, sorry for being slow about responding. I tested it out and it does work, thanks so much. I was talking to Brad about this a few nights ago and he made a practical argument that is often better for us to annotate and filter at the end rather than filter during the calling step. The reasoning is that sometimes there will be variants people are actually interested in that end up getting filtered out because they are in gnomAD and what not, and then they can't recover them if we filter while calling since they never got called in the first place. I think that is a pretty practical take on how to handle these things strategically.

You mentioned in the thread that you had to add this and then aggressively filter the mutect2 tumor-only variants to get to a set that looked reasonable. Do you think you could share what you did? I think it would be a good idea to add in those filters to bcbio, and we could handle gnomAD filtering at that step. Also @ohofmann shared with me sets of filters they have in place for tumor-normal samples that they use to filter their variants. It would be really nice for this all to reside in one place as a downstream filtering step for all of the variant callers. What do you think about that plan? @ohofmann can I share what you are doing with Luca so we can all figure out the best way t do this together?

violethaze74 (approved):
approved

lbeltrame:
We have phased out bcbio in our research and we are no longer using it for anything. This PR has been open for years with no real action, so it's getting closed.

PR #2979, opened 2019-10-15
added new yaml option: use_lowfreq_filter: false to let vardict pick …

naumenko-sa:
…up low frequency variants in high coverage panels

### DIFF
--- a/docs/contents/internals.rst
+++ b/docs/contents/internals.rst
@@ -30,3 +30,190 @@ reference calls in the output variant file.
    :alt: Parallel approaches
 
    Overview of cluster types during parallel execution
+
+Somatic tumor only variant calling pipeline with UMIs step by step
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Minor steps (like tabix'ing of vcfs, indexing of bams) and details (full paths)
+are omitted.
+
+bcbio.yaml config::
+
+    details:
+    - algorithm:
+        aligner: bwa
+        trim_ends: [2,0,2,0]
+        min_allele_fraction: 0.01
+        correct_umis: /path/umi.whitelist.txt
+        tools_off:
+        - gemini
+        umi_type: fastq_name
+        variantcaller:
+        - vardict
+        coverage: target.bed
+        variant_regions: target.bed
+      analysis: variant2
+      description: samplex
+      files:
+      - /path/samplex_1.fq.gz
+      - /path/samplex_2.fq.gz
+      genome_build: hg38
+      metadata:
+        phenotype: tumor
+    resources:
+      fgbio:
+        options: [--min-reads, 3]
+
+- step 1: trimming 2p from reads: trim_ends: [2,0,2,0], indexing::
+
+    bgzip --threads 8 -c <(seqtk trimfq -b 2 -e 0 samplex_1.fq.gz) > samplex_1.fq.gz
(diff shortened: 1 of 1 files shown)


bcbio/ngsalign/postalign.py
```diff
@@ -198,14 +200,16 @@ def correct_umis(data):
...
-           "-i {input_bam} -o {output_bam}")
-    do.run(cmd.format(**locals()), "Correcting UMIs")
+
+    if not utils.file_exists(output_bam):
+        umi_method, umi_tag = _check_umi_type(input_bam)
+        cmd = ("unset JAVA_HOME && "
+               "fgbio {jvm_opts} {io_opts} CorrectUmis "
+               "-t {umi_tag} -m 3 -d 1 -x "
+               "-U {umis_whitelist} "
+               "-i {input_bam} -o /dev/stdout | samtools view -bh > {output_bam}")
```
roryk: Thanks Sergey! It would be good to do a `config_utils.get_program('fgbio", data)` and `config_utils.get_program("samtools", data)` here to make sure we are picking up the bcbio-installed versions. See https://github.com/bcbio/bcbio-nextgen/blob/bd4a01005f8041cb5c876f287eeb30bb3fb0ae9c/bcbio/chipseq/__init__.py#L51-L56 for an example use.

roryk: There are parts in the code where we don't do this, and we've been slowly trying to fix them all, so if you happen to see them when you are doing other stuff, fixing it would be great.

bcbio/ngsalign/postalign.py
```diff
@@ -219,6 +223,7 @@ def umi_consensus(data):
     if not utils.file_uptodate(f1_out, align_bam):
         with file_transaction(data, f1_out, f2_out) as (tx_f1_out, tx_f2_out):
             jvm_opts = _get_fgbio_jvm_opts(data, os.path.dirname(tx_f1_out), 2)
+            print("umi_consensus"+jvm_opts)
```
roryk: Look like a stray debugging statement got left in.

bcbio/ngsalign/postalign.py
```diff
@@ -189,6 +189,8 @@ def correct_umis(data):
     """Correct umis against the whitelist in correct_umi_file
 
     http://fulcrumgenomics.github.io/fgbio/tools/latest/CorrectUmis.html
+
+    we need to sort the output, otherwise it is 300G
```
roryk: I think this looks like a stray note.

bcbio/variation/vardict.py
```diff
@@ -170,13 +170,17 @@ def _run_vardict_caller(align_bams, items, ref_file, assoc_files,
...
+                    lowfreq_filter = " | "
+                else:
+                    lowfreq_filter = " | " + _lowfreq_linear_filter(0, False) + " | "
                 cmd = ("{setup}{jvm_opts}{vardict} -G {ref_file} "
                        "-N {sample} -b {bamfile} {opts} "
                        "| teststrandbias.R "
                        "| var2vcf_valid.pl -A -N {sample} -E {var2vcf_opts} "
-                       "| {contig_cl} | bcftools filter -i 'QUAL >= 0' | {lowfreq_filter} "
-                       "| {fix_ambig_ref} | {fix_ambig_alt} | {remove_dup} | {vcfstreamsort} {compress_cmd}")
+                       "| {contig_cl} | bcftools filter -i 'QUAL >= 0' {lowfreq_filter} "
```
roryk: This is an example of an area where we should try to add the `config_utils.get_program`; it should work for bcftools and hopefully the same for `teststrandbias.R`.

PR #2981, opened 2019-10-16
added config/templates/somatic-duplex-umi.yaml template for Duplex Um…

naumenko-sa:
…i analysis

added description of use_lowfreq_filter to documentation


roryk:
Awesome, thanks for doing all of this Sergey!

Issue #2991, opened 2019-10-23
VEP installation: switch to download indexed cache


lbeltrame:
Apparently something changed in VEP, and now the `-u` flag passed by bcbio isn't actually honored anymore. That means that the installer looks for the tarball in `<vep_cache_dir>/tmp` rather than the path bcbio gives. 

In other words, it proceeded to *redownload* a 15G tarball. I found out because I had to download it separately since Ensembl throttles the ISP used by institution, and now I found (12 hours later!) that the tarball was being downloaded again.

lbeltrame:
From a cursory glance at the VEP code, it looks like it *should* work. But for some reason mine didn't, even though the downloaded tarball was there.

lbeltrame:
And the reason is that `vep_install` now gets the pre-indexed cache, hence the file is different from before, and it doesn't match what VEP expects it to be. 

I guess files should be downloaded from `ftp://ftp.ensembl.org/pub/release-98/variation/indexed_vep_cache/` now.

lbeltrame:
It's also a bug in VEP, which can't allow overriding the indexed cache directory: 

https://github.com/Ensembl/ensembl-vep/issues/628

lbeltrame:
Note that this won't be fixed completely until the VEP bug is fixed.

roryk:
Thanks Luca for being on top of this, especially for working with the folks upstream.

naumenko-sa:
Hi @lbeltrame !

Has it been fixed in VEP? We did not run VEP lately.

SN

PR #2992, opened 2019-10-23
fixed VEP installation path pointing to indexed_vep_cache in v.98

naumenko-sa:
to fix https://github.com/bcbio/bcbio-nextgen/issues/2991


roryk:
The RNA-seq test failed just due to timing out-- this shouldn't affect it so all good.

PR #2996, opened 2019-10-28
Feature request: Can we make bowtie2 `[XS] == null` filter optional?

mjsteinbaugh:
Hi @roryk and @naumenko-sa,

I'd like your input on a bowtie2 setting we've found doesn't always perform well for ChIP-seq, particularly for samples with large global changes in chromatin state.

Here's the relevant line:
https://github.com/bcbio/bcbio-nextgen/blob/master/bcbio/ngsalign/bowtie2.py#L91

```py
base_filter = '-F "[XS] == null and not unmapped {paired_filter}" '
```

For most ChIP-seq samples, the `[XS] == null` filter works well and should be enabled by default. However, we've found for some samples that this doesn't seem to be approriate, and I was wondering if we can add an optional YAML flag to disable this filter.

If you take a look at the code in my forked CPI branch, you'll see we're running this for some samples:

```py
base_filter = '-F "not unmapped {paired_filter}" '
```

What's the best practice for adding a new YAML flag in the bcbio configuration? I can add that and update the pull request.

See also the [Alignment and filtering](https://hbctraining.github.io/Intro-to-ChIPseq/lessons/03_align_and_filtering.html) HBC tutorial, which does a nice job of explaining this step with bowtie2.

Best,
Mike


roryk:
Hi Mike,

We are on the same wavelength-- in the development version there are two options `keep_multimapped` and `keep_duplicates` which control if you keep duplicates or multimapped reads in ChIP/ATAC-seq. Will that do what you wanted?

mjsteinbaugh:
Oh cool, I'll check. Thanks @roryk

mjsteinbaugh:
OK so I'm seeing them in `datadict.py` and `keep_duplicates` is documented (but not `keep_multimapped` yet) in `configuration.rst`, but I'm not seeing the flags implemented in the Python
code.

It looks like we need to add some if/else statements still, unless I'm missing something.

Relevant searches:
- https://github.com/bcbio/bcbio-nextgen/search?q=keep_duplicates&unscoped_q=keep_duplicates
- https://github.com/bcbio/bcbio-nextgen/search?q=keep_multimapped&unscoped_q=keep_multimapped

roryk:
Sorry, you are right, I haven't merged my branch in. I will do that now.

roryk:
https://github.com/bcbio/bcbio-nextgen/blob/master/bcbio/chipseq/__init__.py the code to do it is in here. Should be all set to test out now.

mjsteinbaugh:
Boom, I'll patch my development install here and try it out this week. Awesome.

roryk:
Great, thanks for the p/r Mike, that is awesome.

PR #2998, opened 2019-10-31
fixed a bug in variation/effects/vep

naumenko-sa:
https://github.com/bcbio/bcbio-nextgen/issues/2993


chapmanb:
Thanks much for the quick fix.

PR #3058, opened 2020-01-15
getting started documentation update

naumenko-sa:


roryk:
Thank you, this is so helpful!

PR #3064, opened 2020-01-20
bump vep resources

matthdsm:


matthdsm:
Thanks for merging!
Could you also look into merging https://github.com/chapmanb/cloudbiolinux/pull/332, since the two go together.

Cheers
M

PR #3070, opened 2020-01-23
document noalt_calling

naumenko-sa:


roryk:
These documentation updates are awesome, thanks so much!

PR #3089, opened 2020-02-12
documented bcbio profiling

naumenko-sa:


docs/contents/internals.rst
```diff
@@ -302,13 +302,56 @@ bcbio.yaml config::
...
+usage of bcbio, especially when running  on a server or AWS instance.
+Sometimes running a bcbio project with 32 cores is just 10% more efficient
+than with 16 cores, because a particular configuration might have memory or
+IO related bottlenecks.
+
+- step 1. install and start sysstat deamon: http://www.leonardoborda.com/blog/how-to-configure-sysstatsar-on-ubuntudebian/.
+- step 2. Create a cron job to gather system statistics every minute or two.
+- step 3. Before bcbio start, drop system memory caches. Otherwise memory usage statistic might be misleading::
+
+    sudo su
```
lbeltrame: `sudo -i` does the same thing. Also, I would add a comment like "# or become root" (not everyone uses Ubuntu or Ubuntu derivatives).

docs/contents/internals.rst
```diff
@@ -302,13 +302,56 @@ bcbio.yaml config::
...
+
+    # CPU load
+    sar -q -s $start -e $end |  awk '{print $1","$4}'  | sed 1d | sed 1d > cpu.csv
+    # memory
+    sar -r -s $start -e $end | awk '{print $5}' | sed 1d | sed 1d > mem.csv
+    # IO
+    sar -b -s $start -e $end | awk '{print $5","$6}' | sed 1d | sed 1d > io.csv
+    paste -d "," cpu.csv mem.csv io.csv > usage.csv
+
+    # Example of usage.csv, man sar for more options
```
lbeltrame: Change to "refer to the sar man page for explanations of the fields" (which I assume it has).

docs/contents/internals.rst
```diff
@@ -302,13 +302,56 @@ bcbio.yaml config::
...
+    23:09:01,4.34,24.28,208650.45,26270.58
+    23:10:01,11.09,25.67,0.13,15.46
+    23:11:01,13.56,27.00,4.27,21.99
+    23:12:01,15.44,29.63,26.52,2749.22
+    23:13:01,15.16,29.75,42.93,27.06
+    23:14:01,16.94,30.54,205.26,2740.95
+    23:15:01,15.76,30.57,28.92,2751.62
+    23:16:01,15.77,30.88,6.13,33.59
+
+- step 6. Overlap profiling results with bcbio-nextgen-commands.log to investigate the
```
lbeltrame: An example, even if fictitious, might help there, or perhaps just expand a bit, like "(for example, tracking I/O usage during variant calling)".

docs/contents/internals.rst
naumenko-sa: thanks @lbeltrame , I've updated the page

Issue #3101, opened 2020-02-19
Missing bx-python package


hackdna:
bcbio v1.2.0

[bx-python](https://github.com/bxlab/bx-python) is required by [plink_to_vcf.py](https://github.com/bcbio/bcbio-nextgen/blob/v1.2.0/scripts/utils/plink_to_vcf.py#L26). However:
```
vagrant@bcbio:/vagrant/bcbio-nextgen$ python
Python 3.6.7 | packaged by conda-forge | (default, Feb 28 2019, 09:07:38)
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from bx.seq import twobit
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'bx'
```

bx-python should be included during bcbio installation

hackdna:
@roryk another solution is to remove this script:
* it [isn't a core part of bcbio](https://github.com/chapmanb/cloudbiolinux/pull/339#issuecomment-589708378)
* it [hasn't been updated in years](https://github.com/bcbio/bcbio-nextgen/blame/9289bdf1a48adab55b8ce51058d4bdc7f5d81f97/scripts/utils/plink_to_vcf.py)
* it actually couldn't run at all after upgrade to Python 3
* [the last release of PLINK was almost six years ago](http://zzz.bwh.harvard.edu/plink/)
* deleting unnecessary code reduces [software rot](https://en.wikipedia.org/wiki/Software_rot)

What do you think?

roryk:
It's good to keep it in, IMO-- it's occasionally useful when working with plink files, I've used it several times when we get data in that format. PLINK is a pretty common format for people doing GWAS stuff.

hackdna:
OK, how about moving it (and other similar scripts) to a separate repo then (e.g., bcbio-utils)? These utils then could be made into a PyPI package for easy installation along with dependencies.

roryk:
Thanks, I think we can just leave it rather than creating a new thing. It is some bitrot but its outside of bcbio so the rot is limited. If we start accumulating a bunch of scripts like that then we can think about whether or not it makes sense to have a separate thing, IMO.

hackdna:
The script is part of the repo and the rot tends to spread because then it's unclear what is supposed to be broken and what isn't :) It looks like there are quite a few of these scripts: https://github.com/bcbio/bcbio-nextgen/tree/master/scripts/utils

[...]

PR #3102, opened 2020-02-19
Clean up plink_to_vcf.py and make it Python 3 compatible

hackdna:
See also #3101


roryk:
Thank you! This is more a companion script and not actually part of bcbio, but it is helpful to have it working.

PR #3104, opened 2020-02-21
improve error logging

matthdsm:


matthdsm:
sure, I like those better too!
M

roryk:
Thank you so much!

PR #3118, opened 2020-03-05
improve parallelization of joint calling

matthdsm:
Since joint calling is inherently single threaded, let the scheduler only request one core and submit more jobs.


roryk:
Thanks Matthias. @chapmanb does this look right to you? It would take me a while to look through the code to see if it really is single-threaded or not; do you know for sure?

matthdsm:
We've been running this in production for a few weeks now, without any issues. We've seen some extra thread usage from the JVM, but other than that, no problems here.

Cheers
M

roryk:
Thanks Matthias, sounds good to me. Thank you so much for all of your contributions lately, it is so helpful.

matthdsm:
with pleasure!
M

Issue #3112, opened 2020-02-26
Missing miRBase annotation when installing BDGP6 and dm3 genomes with smallrna datatarget


DrHogart:
Hi,
running
`bcbio_nextgen.py upgrade --genomes BDGP6 --datatarget smallrna`
results to the
`List of genomes to get (from the config file at '{'genomes': [{'dbkey': 'BDGP6', 'name': 'D melangogaster (BDGP6)', 'indexes': ['seq'], 'annotations': ['transcripts']}], 'genome_indexes': ['bwa', 'bowtie2', 'rtg', 'star'], 'install_liftover': False, 'install_uniref': False}'): D melangogaster (BDGP6)`.
As you see there are no miRBase annotation files, only 'transcripts'. The same with dm3 genome. Correspondingly, there was no srnaseq folder after upgrading. So, smallRNA-seq analysis doesn't work. At the same time upgrade with smallrna datatarget for hg19 gets miRBase annotations correctly.

bcbio 1.2.0

Could you please add the srnaseq data in the BDGP6 genome resources?

naumenko-sa:
Hi Sergei @DrHogart!

Thanks for reporting the issue!

While miRNA is mostly supported for human genome, 
I was able to run _A.thaliana_ miRNA analysis once: https://github.com/bcbio/bcbio-nextgen/issues/1416

To push this analysis for Drosophila, we need to create a mirbase recipe in cloudbiolinux:
https://github.com/chapmanb/cloudbiolinux/tree/master/ggd-recipes/BDGP6
similarly to hg19 for human
https://github.com/chapmanb/cloudbiolinux/blob/master/ggd-recipes/hg19/mirbase.yaml
or A.thaliana:
https://github.com/chapmanb/cloudbiolinux/blob/master/ggd-recipes/TAIR10/mirbase.yaml

And then update resources.yaml in bcbio:
https://github.com/bcbio/bcbio-nextgen/blob/master/config/genomes/BDGP6-resources.yaml
similarly to A.thaliana
https://github.com/bcbio/bcbio-nextgen/blob/master/config/genomes/TAIR10-resources.yaml

If you could help creating and testing this recipe (making sure all files and downloaded and they correspond to the correct reference) with a pull request (PR), that would speed up the process.

Here is how to test a recipe:
https://github.com/chapmanb/cloudbiolinux/blob/master/doc/hacking.md#testing-a-ggd-recipe

Sergey

[...]

PR #3119, opened 2020-03-05
Fix indentations in srnaseq resources of BDGP6

DrHogart:
Fixing the bug indentations preventing the execution of smallRNA-seq analysis for BDGP6 (see [this comment of issue #3112 ](https://github.com/bcbio/bcbio-nextgen/issues/3112#issuecomment-594697863))


naumenko-sa:
Thanks for fixing this!

PR #3131, opened 2020-03-18
Change raw.github.com URLs to raw.githubusercontent.com

hackdna:
raw.github.com permanently redirects to raw.githubusercontent.com


naumenko-sa:
Thanks! I think it is because of:
https://developer.github.com/changes/2014-04-25-user-content-security/

PR #3140, opened 2020-03-20
fix duplicated rows in the multiqc report

### LINKED ISSUE #3139: Duplicated rows in multiqc report
Hi everyone!

This bug was observed in a typical project:
https://github.com/bcbio/bcbio-nextgen/blob/master/config/templates/somatic-single-umi.yaml

Multiqc creates two rows per sample in 
`bcbio_run/final/project/multiqc/multiqc_Data/multiqc_general_stats.txt` table.
It seems that the second row is from mosdepth, named sample-coverage with only two columns filled: >=30X and median. And these columns are empty for sample row.

Analysis from @vladsaveliev :
The reason it happens is because mosdepth output file format looks like <prefix>.mosdepth.region.dist.txt, and mosdepth MultiQC module assumes that <prefix> is the sample name (https://github.com/ewels/MultiQC/blob/281fd5d64a0a4af9ba25b83969b4d11c0d611028/multiqc/utils/search_patterns.yaml#L360-L363) and grabs it as one. 

However, mosdepth files bcbio creates are named like <sample>-coverage.mosdepth.region.dist.txt, with this extra "-coverage" suffix, so when MultiQC finds these files, it adds data for a sample named <sample>-coverage.It wasn't an issue before because mosdepth module is new in MultiQC, and bcbio assumes it doesn't exists and does all coverage calculation on its own in MultiQC_bcbio: https://github.com/MultiQC/MultiQC_bcbio/blob/master/multiqc_bcbio/bcbio.py#L358.

So there are 3 ways to fix it:
- disable mosdepth module by adding -e mosdepth to the multiqc command line in bcbio.
- rename <sample>-coverage.mosdepth.region.dist.txt to  <sample>.mosdepth.region.dist.txt in bcbio.
- or add these 3 lines into multiqc config to make mosdepth parse the sample name correctly: https://github.com/umccr/umccrise/blob/master/umccrise/multiqc/multiqc_config.yaml#L232-L234 and fully switch to mosdepth MultiQC module instead of the coverage calculation in MultiQC_bcbio.
```

naumenko-sa:
close #3139 
thanks @vladsaveliev!

### DIFF
--- a/bcbio/qc/multiqc.py
+++ b/bcbio/qc/multiqc.py
@@ -300,12 +300,14 @@ def _create_list_file(paths, out_file):
     return out_file
 
 def _create_config_file(out_dir, samples):
-    """Provide configuration file hiding duplicate columns.
+    """Provide configuration file for multiqc report."""
 
-    Future entry point for providing top level configuration of output reports.
-    """
     out_file = os.path.join(out_dir, "multiqc_config.yaml")
     out = {"table_columns_visible": dict()}
+
+    extra_fn_clean_trim = []
+    extra_fn_clean_trim.extend(["coverage.mosdepth.region.dist", "coverage.mosdepth.global.dist"])
+    out["extra_fn_clean_trim"] = extra_fn_clean_trim
 
     # Avoid duplicated bcbio columns with qualimap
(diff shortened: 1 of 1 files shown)


roryk:
Thanks for fixing this!

PR #3144, opened 2020-03-23
dockerfile tweaks

matthdsm:
I've removed a bunch of stuff that seemend superflous and tried to streamline the build.
For maximum perfomance on automated builds, I think we should move to a bcbio base image with all dependencies installed and base the automated build on this image.

That way, the auto-build could just `pip install` the latest bcbio codebase into the image and start running tests.
Downside here is that the base image would have to be built and pushed manually.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,66 +1,43 @@
-FROM stackbrew/ubuntu:14.04
-MAINTAINER Brad Chapman "https://github.com/chapmanb"
+FROM debian:stretch-slim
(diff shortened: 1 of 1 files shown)


matthdsm:
Is the dockerfile in this repo still used then? If not, I think it should be removed to avoid confusion.

M

roryk:
Thanks, I think we aren't using the Dockerfile here. I'll remove it.

Issue #3148, opened 2020-03-25
CalledProcessError in test_4_cwl_rnaseq


hackdna:
**Version info**
 - bcbio version (bcbio_nextgen.py --version): 1.2.0 (a0db761)
 - OS name and version: Ubuntu Bionic

**To Reproduce**
 - Exact bcbio command you have used: `py.test -p no:cacheprovider -p no:stepwise tests/bcbio_vm -v -s -m cwl_docker_rnaseq`
 - Your sample configuration file: N/A

**Observed behavior**
```
=================================== FAILURES ===================================
__________________________ TestCWL.test_4_cwl_rnaseq ___________________________

self = <test_docker.TestCWL object at 0x7f34f77cfa50>, install_test_files = None
data_dir = '/home/travis/build/bcbio/bcbio-nextgen/tests/data/automated'

    @pytest.mark.cwl
    @pytest.mark.cwl_docker_rnaseq
    def test_4_cwl_rnaseq(self, install_test_files, data_dir):
        """CWL: run a RNA-seq workflow using Docker.
        """
        with install_cwl_test_files(data_dir) as workdir:
            with utils.chdir(os.path.join(workdir, "rnaseq")):
                cl = ["bash", "./run_generate_cwl.sh"]
                if os.path.exists("cromwell_work"):
                    shutil.rmtree("cromwell_work")
                cl = ["bcbio_vm.py", "cwlrun", "cromwell", "rnaseq-workflow"]
>               subprocess.check_call(cl)

tests/bcbio_vm/test_docker.py:99: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

popenargs = (['bcbio_vm.py', 'cwlrun', 'cromwell', 'rnaseq-workflow'],)
kwargs = {}, retcode = 1
cmd = ['bcbio_vm.py', 'cwlrun', 'cromwell', 'rnaseq-workflow']

    def check_call(*popenargs, **kwargs):
        """Run command with arguments.  Wait for command to complete.  If
        the exit code was zero then return, otherwise raise
        CalledProcessError.  The CalledProcessError object will have the
        return code in the returncode attribute.
    
        The arguments are the same as for the call function.  Example:
    
        check_call(["ls", "-l"])
        """
        retcode = call(*popenargs, **kwargs)
        if retcode:
[...]

PR #3156, opened 2020-03-31
Remove unused pytest fixture workdir

hackdna:


roryk:
Looks like we can nuke this line from travis? https://github.com/bcbio/bcbio-nextgen/blob/76dc03337c86e376baccfded99e8dd92b8c5e555/.travis.yml#L29

hackdna:
Yes, let me see if that fixes the build. This line was introduced in https://github.com/bcbio/bcbio-nextgen/commit/233c45834006ab955506a30981e943b07c6a665b but not clear as to why from that comment.

hackdna:
Looks like that did it. The build is still failing but that's due to #3148.

PR #3163, opened 2020-04-03
Fix generation of PureCN summary output

lbeltrame:
The current code did not check if the copy number values were valid
(some can be NA) and thus it would error out when trying to coerce them
with a float.

The easiest solution, since bcbio already depends on pandas, was to
ditch the custom parsing and use pandas to read the file, and skip lines
with missing values.

As a micro optimization, reading the LOH file was put outside the loop,
to reduce needless I/O.

### DIFF
--- a/bcbio/heterogeneity/loh.py
+++ b/bcbio/heterogeneity/loh.py
@@ -8,6 +8,7 @@ import os
 import decimal
 import uuid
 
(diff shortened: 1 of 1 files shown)


roryk:
Thank you!! This is a lot nicer.

Issue #3191, opened 2020-04-15
Make installation and upgrading more robust


hackdna:
Issues often arise due to runtime errors during installation or upgrading of bcbio, for example: #3165, #3169, #3180, etc. Other issues arise due to coding errors, for example: #3168, #3188, etc.

In many cases error messages don't provide enough information to the users, for example: "subprocess.CalledProcessError: Command returned non-zero exit status 1".

Issues due to runtime errors can be reduced by better error handling that provides more informative error messages. Issues due to coding errors can be reduced by running the installer as part of Travis CI builds or at least adding that as an additional step of the release process.

amizeranschi:
@hackdna I've just noticed you're looking into the installer issues. FWIW, the installer is broken for the current stable release. Please also see: https://github.com/bcbio/bcbio-nextgen/issues/3223

hackdna:
@amizeranschi thanks for heads up. Looks like @naumenko-sa is helping you figure it out.

amizeranschi:
Yes, the issue is now solved. Many thanks to you guys for working on bcbio-nextgen!

PR #3208, opened 2020-04-30
Conda install cleanup

hackdna:
Toward #3191:
* Configure Conda channels via .condarc configuration file
* Remove _get_conda_channels() and install_conda_build()
* PEP8 fixes


roryk:
Thank you! Good find with letting users override this themselves.

PR #3214, opened 2020-05-11
dont use HaplotypeCallerSpark for RNA-seq variant calling

naumenko-sa:


roryk:
Thanks for figuring out what was happening here Sergey!

PR #3222, opened 2020-05-15
Update Travis CI badge to track builds on the master branch only

hackdna:


roryk:
Thanks for fixing this!

Issue #3240, opened 2020-05-27
unable to do fresh install, conda problem


parlar:
**Version info**
 - bcbio version (bcbio_nextgen.py --version): latest stable (1.2.3)
 - OS name and version: Ubuntu 18.04

**To Reproduce**
 - Exact bcbio command you have used: 
python bcbio_nextgen_install.py /home/parlar/bcbio --tooldir=/home/user/bcbio/tools --nodata

**Observed behavior**

Traceback (most recent call last):
  File "/home/parlar/bcbio/anaconda/bin/conda", line 12, in <module>
    from conda.cli import main
ModuleNotFoundError: No module named 'conda'
Traceback (most recent call last):
  File "bcbio_nextgen_install.py", line 290, in <module>
    main(parser.parse_args(), sys.argv[1:])
  File "bcbio_nextgen_install.py", line 47, in main
    bcbio = install_conda_pkgs(anaconda, args)
  File "bcbio_nextgen_install.py", line 108, in install_conda_pkgs
    "--file", os.path.basename(REMOTES["requirements"]), TARGETPY], env=env)
  File "/usr/lib/python2.7/subprocess.py", line 190, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/home/parlar/bcbio/anaconda/bin/conda', 'install', '--yes', '--file', 'requirements-conda.txt', 'python=3.6']' returned non-zero exit status 1

**Expected behavior**
Normal, clean installation

**Log files**
NA

**Additional context**
Removed potentially conflicting conda installations, tried 'development' instead of 'stable' with the same result.

parlar:
In the installer, the following line in the ```install_conda_pkgs``` function seems to break downstream use of conda and mamba:

```
    subprocess.check_call([mamba_bin, "install", "--yes", "--only-deps", "bcbio-nextgen",
                           TARGETPY], env=env)
```

naumenko-sa:
Hi Pär @parlar!

Thank for reporting - sorry about the issue!
I confirm I see it as well in my AWS Ubuntu install.

I tried to revert changes we recently did last month while cleaning the installation scripts,
but the error is still there, and we also had successful installations when testing the changes and last week.

So it is likely that the issue is not with the installation script, maybe it is rather related to https://github.com/conda/conda/issues/9672 and similar conda issues?

[...]

PR #3246, opened 2020-05-28
Prevent silent updates of Miniconda during installation

hackdna:
* Specify exact Miniconda version in the installer to prevent issues like #3240 
* Simplify installer code (toward #3191)
* Update Conda repo URL


roryk:
Thanks! This is a good  idea.

Issue #3176, opened 2020-04-07
Move Docker tests out of the bcbio-nextgen repo


hackdna:
[Docker tests](https://github.com/bcbio/bcbio-nextgen/blob/v1.2.3/tests/bcbio_vm/test_docker.py) are exercising `bcbio_vm.py` which is a tool from [bcbio-nextgen-vm repo](https://github.com/bcbio/bcbio-nextgen-vm) and are dependent on Docker containers that are built in [bcbio_docker repo](https://github.com/bcbio/bcbio_docker).

On the other hand, actual unit and integration tests for bcbio are not running in Travis CI build of `bcbio-nextgen` repo. Also, `bcbio-nextgen-vm` has no test coverage.

Docker tests should be moved either to `bcbio-nextgen-vm` or `bcbio_docker` repo while `bcbio-nextgen` unit and integration tests should be added to `bcbio-nextgen` Travis CI build which would provide more test coverage and require less running time (run all pipelines with subsampled input files?)

Also, this would allow running the installer as part of Travis builds.

PR #3248, opened 2020-05-28
Clean up CI configuration, installer and tests

hackdna:
Some initial work toward #3176 and also #3191.


roryk:
ChIP-seq and ATAC-seq tests should be passing though, I will fix those.

PR #3266, opened 2020-06-12
Corrected the omission of *-resources.yaml

a113n:
resources.yaml is required to be included as secondary files in order to run the postprocess_alignment_to_rec and prep_samples_to_rec steps


roryk:
Thank you so much for this fix! Sorry for the delay in getting to it.

PR #3273, opened 2020-06-22
WGBS pipeline improvements

kmendler:
- Allow TrimGalore to use more than one thread
- Allow Bismark to use more than 2 threads
- Add directionality to Kit object 
- Update docs to include changes from this PR and Sergey's previous commits

The pipeline now runs more efficiently on machines with more resources and can be configured with any supported wgbs kit or manually through the yaml (see example in the updated docs).


naumenko-sa:
Thanks @kmendler !

roryk:
Thanks Kerrin! This is so awesome.

PR #3290, opened 2020-07-01
Resource requirements

hackdna:
Update requirements for tool installation and data upgrade.

### DIFF
--- a/docs/contents/installation.md
+++ b/docs/contents/installation.md
@@ -24,7 +24,7 @@ or inside your home directory:
 python bcbio_nextgen_install.py /home/user/bcbio --tooldir=/home/user/bcbio/tools --nodata 
 ```
 
-Installation takes ~30 min.
+Installation takes 30 minutes or more (depending on the speed of your storage and internet connection). Recommended HPC job parameters for the installation process: 1 CPU core, 2GB memory, 1 hour run time.
 
 Check if installation works:
 ```
--- a/docs/contents/installation.md
+++ b/docs/contents/installation.md
@@ -224,7 +224,7 @@ Note that muTect does not provide an easy way to query for the current version, 
 
 bcbio-nextgen provides a wrapper around external tools and data, so the actual tools used drive the system requirements. For small projects, it should install on workstations or laptops with a couple GB of memory, and then scale as needed on clusters or multicore machines.
 
-Disk space requirements for the tools, including all system packages are under 4GB. Biological data requirements will depend on the genomes and aligner indices used, but a suggested install with GRCh37 and bowtie/bwa2 indexes uses approximately 35GB of storage during preparation and ~25GB after:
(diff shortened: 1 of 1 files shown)


docs/contents/installation.md
```diff
@@ -163,6 +163,7 @@ Tune the upgrade with these options:
 * Leave out the `--tools` option if you don't want to upgrade third party tools. If using `--tools`, it will use the same directory as specified during installation. If you're using an older version that has not yet gone through a successful upgrade or installation and saved the tool directory, you should manually specify `--tooldir` for the first upgrade. You can also pass `--tooldir` to install to a different directory.
 * Leave out the `--data` option if you don't want to get any upgrades of associated genome data.
 * Some aligners such as STAR don't have pre-built indices due to the large file sizes of these. You set the number of cores to use for indexing with `--cores 8`.
+* For example, recommended HPC job parameters for `bcbio_nextgen.py upgrade -u skip --data --datatarget rnaseq --genomes GRCh37` are: 2 CPU cores, 2GB memory, and 2 hours run time.
```
lbeltrame, line 166: I don't remember if it is still the case, but does `--datatarget rnaseq` still require the user to generate the HISAT2 index? If so, the memory requirements are far too low.

docs/contents/installation.md
```diff
@@ -224,7 +225,7 @@ Note that muTect does not provide an easy way to query for the current version,
 
 bcbio-nextgen provides a wrapper around external tools and data, so the actual tools used drive the system requirements. For small projects, it should install on workstations or laptops with a couple GB of memory, and then scale as needed on clusters or multicore machines.
 
-Disk space requirements for the tools, including all system packages are under 4GB. Biological data requirements will depend on the genomes and aligner indices used, but a suggested install with GRCh37 and bowtie/bwa2 indexes uses approximately 35GB of storage during preparation and ~25GB after:
+Disk space requirement for the tools, including all system packages is about 22GB (or more, depending on the type of the file system). Biological data requirements will depend on the genomes and aligner indices used, but a suggested install with GRCh37 and bowtie/bwa2 indexes uses approximately 35GB of storage during preparation and ~25GB after:
```
lbeltrame, line 228: Does this 22GB figure include also the data? As far as I can read from the old text, it only mentioned the actual programs without any data files (but I'm not sure if it is still the same now).

docs/contents/installation.md
hackdna: I ran that command in a SLURM job using the given parameters and it finished successfully.

docs/contents/installation.md
hackdna: This is after `python bcbio_nextgen_install.py [bcbio_path] --tooldir=[bcbio_tools_path] --nodata`.

docs/contents/installation.md
naumenko-sa: it does not generate the index, to generate index it is `--aligners hisat2` and yes, it will need more RAM

PR #3302, opened 2020-07-14
Add option to preserve test output dir between test runs

### LINKED ISSUE #3155: Avoid editing the source code to keep the work directory between test runs
The current behavior (v1.2.1) is to delete the work directory before a test run. While it is useful to keep the work directory in some circumstances it should not be necessary to edit the source code to accomplish that (https://github.com/bcbio/bcbio-nextgen/commit/936ed35121d83f47b868b673a14f6775ed0b076b).

Note: it appears that the work directory is deleted after each successful test run but not after a failed run.

hackdna:
Add `--keep-test-dir` option to close #3155. Simplify test configuration. Move test output directory outside the source tree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.idx
 *.swo
 *.swp
+**/bcbiotx/
 bcbio/pipeline/version.py
 bcbio_nextgen.egg-info/
 build/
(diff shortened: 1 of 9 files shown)


roryk:
Thank you, sorry for taking forever to get to this. It works great and this refactoring is way better. Thanks so much!

PR #3310, opened 2020-07-20
remove samtools sort -n which breaking bismark for some samples

naumenko-sa:


bcbio/wgbsseq/deduplication.py
```diff
@@ -13,7 +13,8 @@ def dedup_bismark(data):
     """
     config = data["config"]
     input_file = datadict.get_work_bam(data)
-    input_file = bam.sort(input_file, config, order="queryname")
+    # don't sort even by read names
```
lbeltrame, line 16: Perhaps telling why sorting isn't a good idea would be useful.

naumenko-sa: Hi @lbeltrame!

some samples fail bismark when you alter the bam in any way
https://github.com/FelixKrueger/Bismark/issues/360

S

Issue #3337, opened 2020-09-03
/bin/bash: line 1: 53677 Segmentation fault


tiger815:
hi 

I'm a beginner for bcbio-nextgen pipeline. 

After installation, I've been attempting to run my WES data for testing.
Unfortunately, I got an error message.

Any help is greatly appreciated.
 
Thanks

Version info
Bcbio Version info : 1.2.3
Ubuntu 18.04 and 128G ram 

tiger815@shdeep:~/NGS/KSO/KSO/work$ bcbio_nextgen.py ../config/KSO.yaml -n 8
Running bcbio version: 1.2.3
global config: /data/home/tiger815/NGS/KSO/KSO/work/bcbio_system.yaml
run info config: /data/home/tiger815/NGS/KSO/KSO/config/KSO.yaml
[2020-09-03T02:35Z] System YAML configuration: /data/bcbio/galaxy/bcbio_system.yaml.
[2020-09-03T02:35Z] Locale set to C.UTF-8.
[2020-09-03T02:35Z] Resource requests: bwa, gatk, sambamba, samtools; memory: 3.00, 3.00, 3.00, 3.00; cores: 8, 8, 8, 8
[2020-09-03T02:35Z] Configuring 1 jobs to run, using 8 cores each with 24.1g of memory reserved for each job
[2020-09-03T02:35Z] Timing: organize samples
[2020-09-03T02:35Z] multiprocessing: organize_samples
[2020-09-03T02:35Z] Using input YAML configuration: /data/home/tiger815/NGS/KSO/KSO/config/KSO.yaml
[2020-09-03T02:35Z] Checking sample YAML configuration: /data/home/tiger815/NGS/KSO/KSO/config/KSO.yaml
[2020-09-03T02:35Z] Retreiving program versions from /data/bcbio/manifest/python-packages.yaml.
[2020-09-03T02:35Z] Retreiving program versions from /data/bcbio/manifest/r-packages.yaml.
[2020-09-03T02:35Z] Testing minimum versions of installed programs
[2020-09-03T02:35Z] Timing: alignment preparation
[2020-09-03T02:35Z] multiprocessing: prep_align_inputs
[2020-09-03T02:35Z] Skipping trimming of KSO.
[2020-09-03T02:35Z] Resource requests: ; memory: 1.00; cores: 1
[2020-09-03T02:35Z] Configuring 2 jobs to run, using 1 cores each with 1.00g of memory reserved for each job
[...]

PR #3338, opened 2020-09-03
increase default RAM

naumenko-sa:
from 3G to 4G / core. save ppl from segfaults


naumenko-sa:
This was surfacing many times, for example here: https://github.com/bcbio/bcbio-nextgen/issues/3337

roryk:
Thank you, good idea!

Issue #3355, opened 2020-10-21
Enabling GVCF file square off when starting from existing RNAseq BAM and GVCF files


WimSpee:
**Version info**
- bcbio version (`bcbio_nextgen.py --version`): 1.1.5

**To Reproduce**
Exact bcbio command you have used:
```
bcbio_nextgen.py {CONFIG_YAML} -t ipython  -n 41 -s sge  -q main.regular -r vf=5G,h_rss=50G'
```
Your sample configuration file:
Repeated for many RNAseq samples. 
```
 'rna-seq variant calling': {
        'algorithm': {
            'aligner': False,
            'strandedness': 'unstranded',
            'transcript_assembler': 'stringtie',
            'variantcaller': False,
            'jointcaller': 'gatk-haplotype-joint',
            'mark_duplicates': False,
            'nomap_split_targets': 1000,
            'ploidy': {'default': 2},
            'realign': False,
            'recalibrate': False,
            'tools_off': ['gemini'],
        },
        'analysis': 'RNA-seq',
        'description': '  SAMPLE_NAME',
        'files': ['input_bam/SAMPLE_NAME-ready.bam'],
        'genome_build': 'REFERENCE_NAME',
        'lane': 'SAMPLE_NAME',
        'metadata': {'batch': 'BATCH_NAME'},
        'vrn_file': 'input_gvcf/SAMPLE_NAME.vcf.gz',
    },
```

**Observed behavior**
Per sample:
* the provided BAM and GVCF are found, STAR and GATK haplotype caller are thus skipped as intended. 
* the raw (*.counts) and normalized counts (*.sf) files are created.  Did not yet try to also provide these files already as input. And thus skip salmon and the program that created the raw counts. 
* the GTF files are made.  Did not yet try to also provide these files already as input. And thus skip Stringtie. 

Over all samples:
- combined raw counts file (combined.counts) is made. 
- combined normalized counts files are made (combined.fpkm,combined.isoform.fpkm )
- No squared of multi-sample VCF is made from the provided GVCF files 

**Expected behavior**
The same as the observed behavior. With the important additions of:

1.  the squared off multi-sample VCF file being made 
2. RNAseq specific soft-filtering of the multi-sample VCF.

The squared off multi-sample VCF file is made when starting with existing BAM and GVCF files for whole genome sequencing data. https://github.com/bcbio/bcbio-nextgen/issues/2336

It would be nice if bcbio could do the same for existing RNAseq GVCF files. 

[...]

PR #3361, opened 2020-10-28
Allow for RNAseq analysis starting from GVCF level

WimSpee:
See #3355

### DIFF
--- a/bcbio/pipeline/rnaseq.py
+++ b/bcbio/pipeline/rnaseq.py
@@ -97,14 +97,15 @@ def rnaseq_variant_calling(samples, run_parallel):
     """
     samples = run_parallel("run_rnaseq_variant_calling", samples)
     variantcaller = dd.get_variantcaller(to_single_data(samples[0]))
-    if variantcaller and ("gatk-haplotype" in variantcaller):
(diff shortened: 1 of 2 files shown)


roryk:
Thank you for doing this, sorry for being slow about commenting. This is great! I am so psyched. I need to test it out but I will definitely merge it in for the next release. I've been wanting this feature for a while, I am so happy you implemented it.

WimSpee:
Hi Rory. Thank you for having a look. I understand that also you might be busy with other things. 

We look forward to bcbio supporting (recurrent) joint analysis from existing (expanding set) of singe sample results more. 
 So I am also very exited about this functionality.  And it looks like not that much needs to be changed how bcbio currently works. 

One thing that still needs to be tested is passing in existing Stringtie, Salmon and featureCounts files. I might be able to do that in the coming days/weeks, and I'll update the pull request if needed. 

For now it is also still possible for us to patch our local bcbio installation with the changes of this pull request
In general though it definitively makes sense to us to have this functionality in bcbio.

Thank you.

roryk:
Thank you! I merged this in, sorry for making you wait so long. I somehow screwed it up and this specific P/R didn't get merged in but the commits from it are. But it should be in there now.

PR #3364, opened 2020-11-03
PureCN analysis support in bcbio:

naumenko-sa:
- prepare purecn bed file
- calculate purecn coverage
- purecn rscript wrappers
- run purecn without gatk-cnv and cnvkit segmentation
- update mutect variant calling when creating SNV PON
- correct batch pairing logic to allow PON processing
- merge SNV pon
- normal db creation and upload to final
- PON creation works, passes cancermulti
- T-only + PON works
- t/n calling + pon works
- standard production pipeline works
- bug fixes

### DIFF
--- a/bcbio/structural/__init__.py
+++ b/bcbio/structural/__init__.py
@@ -33,7 +33,7 @@ _CALLERS = {
 _NEEDS_BACKGROUND = set(["cn.mops"])
 _GLOBAL_BATCHING = set(["seq2c"])
 # CNV callers that have background references
-_CNV_REFERENCE = set(["seq2c", "cnvkit", "gatk-cnv"])
+_CNV_REFERENCE = set(["seq2c", "cnvkit", "gatk-cnv", "purecn"])
 
 def _get_callers(items, stage, special_cases=False):
     """Retrieve available callers for the provided stage.
@@ -54,6 +54,7 @@ def _get_callers(items, stage, special_cases=False):
     return callers
 
 def get_svcallers(data):
+    """why this exists? dd.get_svcaller(data)"""
     svs = data["config"]["algorithm"].get("svcaller")
     if svs is None:
         svs = []
@@ -63,6 +64,7 @@ def get_svcallers(data):
 
 def _handle_multiple_svcallers(data, stage):
     """Retrieve configured structural variation caller, handling multiple.
+       data is one sample
     """
     svs = get_svcallers(data)
     # special cases -- prioritization
@@ -130,6 +132,29 @@ def validate_sv(data):
     """
     return [[validate.evaluate(data)]]
 
+def create_cnv_pon(samples):
+    """create normal db using SNV PON and coverage information"""
+    coverage_files = []
+    sample0 = utils.to_single_data(samples[0])
+    purecn_pon_build = tz.get_in(["config", "algorithm", "purecn_pon_build"], sample0)
+    if purecn_pon_build:
+        for sample in (utils.to_single_data(x) for x in samples):
+            sample_coverage = tz.get_in(["depth", "bins", "purecn"], sample)
+            if os.path.exists(sample_coverage):
+                coverage_files.append(sample_coverage)
+        work_dir = tz.get_in(["dirs", "work"], sample0)
+        out_dir = utils.safe_makedir(os.path.join(work_dir, "gemini"))
+        purecn_coverage_files_txt = os.path.join(out_dir, "purecn_coverage_files.txt")
+        with open(purecn_coverage_files_txt, "w") as out_handle:
+            for fname in coverage_files:
+                out_handle.write(fname + "\n")
+        snv_pon = tz.get_in(["config", "algorithm", "purecn_snv_pon"], sample0)
+        normal_db = os.path.join(out_dir, "normalDB_hg38.rds")
+        # delete normal db in work/gemini if you'd like to rebuild it!
+        if os.path.exists(snv_pon) and len(coverage_files) >=3 and not os.path.exists(normal_db):
+            purecn.create_normal_db(purecn_coverage_files_txt, snv_pon, out_dir)
+    return samples
+
 def batch_for_sv(samples):
     """Prepare a set of samples for parallel structural variant calling.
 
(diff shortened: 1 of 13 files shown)


bcbio/variation/vcfutils.py
```diff
@@ -324,6 +330,25 @@ def _do_merge(orig_files, out_file, config, region):
...
+           "-R", ref_file,
+           "-o", out_vcf,
+           "--minimumN", minimumN]
+    out_gz = out_vcf + ".gz"
+    if not os.path.exists(out_gz):
+        try:
+            cmd_line = " ".join([str(x) for x in cmd]) + " -V " + " -V ".join(orig_files)
+            do.run(cmd_line, "combine variants")
+        except subprocess.CalledProcessError as msg:
+              logger.info("PON merge failed")
```
lbeltrame, line 347: Does this report the error, or just prints that error message? If that's the case, perhaps reporting it would help. Also, AFAICS, this makes errors go on "semi silently": I'm not sure that's what we want.

There's another instance of this somewhere else, same problem.

bcbio/variation/vcfutils.py
```diff
@@ -324,6 +330,25 @@ def _do_merge(orig_files, out_file, config, region):
         bgzip_and_index(out_file, config)
     return out_file
 
+def _do_combine_variants(orig_files, out_vcf, ref_file, config, region, minimumN = 3):
```
lbeltrame, line 333: Is the GATK3 dependency really necessary? This makes matters complicated due to it only supporting ancient Java.

bcbio/structural/__init__.py
```diff
@@ -130,6 +132,29 @@ def validate_sv(data):
...
+                coverage_files.append(sample_coverage)
+        work_dir = tz.get_in(["dirs", "work"], sample0)
+        out_dir = utils.safe_makedir(os.path.join(work_dir, "gemini"))
+        purecn_coverage_files_txt = os.path.join(out_dir, "purecn_coverage_files.txt")
+        with open(purecn_coverage_files_txt, "w") as out_handle:
+            for fname in coverage_files:
+                out_handle.write(fname + "\n")
+        snv_pon = tz.get_in(["config", "algorithm", "purecn_snv_pon"], sample0)
+        normal_db = os.path.join(out_dir, "normalDB_hg38.rds")
+        # delete normal db in work/gemini if you'd like to rebuild it!
```
lbeltrame, line 153: This needs to be documented somewhere (if it has not been done already).

bcbio/structural/purecn.py
```diff
@@ -46,6 +52,63 @@ def run(items):
...
+    cmd = [ rscript, purecn_r,
+            "--out", work_dir,
+            "--tumor", sample_coverage,
+            "--sampleid", sample_name,
+            "--vcf", variants_vcf,
+            "--normaldb", normaldb,
+            "--mappingbiasfile", mappingbiasfile,
+            "--intervals", intervals,
+            "--snpblacklist", simple_repeat_bed,
+            "--genome", "hg38",
```
lbeltrame, line 86: This hardcodes the genome to `hg38`. It should instead use `genome_build` from the configuration.

bcbio/structural/purecn.py
```diff
@@ -240,3 +313,95 @@ def _loh_to_vcf(cur):
...
+    optimized_bed = basename + ".optimized.bed"
+    rscript = utils.Rscript_cmd("r36")
+    interval_file_r = utils.R_package_script("r36", "PureCN", "extdata/IntervalFile.R")
+    ref_file = dd.get_ref_file(data)
+    mappability_resource = dd.get_variation_resources(data)["purecn_mappability"]
+    cmd = [rscript, interval_file_r, "--infile", bed_file,
+          "--fasta", ref_file,
+          "--outfile", ready_file, 
+          "--offtarget",
+          "--genome hg38",
```
lbeltrame, line 338: Same as before. This should respect the user's configuration.

bcbio/structural/purecn.py
```diff
@@ -240,3 +313,95 @@ def _loh_to_vcf(cur):
...
+        cmd = [rscript, coverage_r,
+               "--outdir", work_dir,
+               "--bam", bam_file,
+               "--intervals", intervals]
+        try:
+            cmd_line = "export R_LIBS_USER=%s && %s && %s" % (utils.R_sitelib(env = "r36"),
+                                                              utils.get_R_exports(env = "r36"),
+                                                              " ".join([str(x) for x in cmd]))
+            do.run(cmd_line, "PureCN coverage")
+        except subprocess.CalledProcessError as msg:
```
lbeltrame, line 378: What happens here? (same as other questions) It looks like the error is, in fact, ignored.

bcbio/structural/purecn.py
```diff
@@ -240,3 +313,95 @@ def _loh_to_vcf(cur):
...
+              mapping_bias_hg38.rds
+              normalDB_hg38.rds
+    """
+    rscript = utils.Rscript_cmd("r36")
+    normaldb_r = utils.R_package_script("r36", "PureCN", "extdata/NormalDB.R")
+    cmd = [rscript, normaldb_r,
+           "--outdir", out_dir,
+           "--coveragefiles", coverage_files_txt,
+           "--normal_panel" , snv_pon,
+           "--genome hg38",
```
lbeltrame, line 397: Hardcoded genome (same as other instances).

bcbio/variation/multi.py
```diff
@@ -204,6 +205,7 @@ def _pick_lead_item(items):
         for data in items:
             if vcfutils.get_paired_phenotype(data) == "tumor":
                 return data
+        print(items)
```
lbeltrame, line 208: Leftover debug?

lbeltrame:
Posted https://github.com/bcbio/bcbio-nextgen/pull/3366 to fix the hardcoded genome. 

@naumenko-sa In case you need help with PureCN related stuff feel free to poke me if need be.

bcbio/variation/vcfutils.py
naumenko-sa: Agree, I was going to get rid of gatk3 dependency as well. Unfortunately, it turned out not to be very straightforward (see purecn github - there was a discussion on that). what gatk3 is doing here is not supported directly by bcftools or gatk4/picard. It is just (1) merging vcfs into a multisample vcf, (2) filtering sites having at least 3 calls (in any of the samples). (2) is an issue. For example, it is supported in bcftools with plugins, but plugins don't get installed via conda: https://github.com/bioconda/bioconda-recipes/issues/23993
gatk3 goes open-source https://github.com/bioconda/bioconda-recipes/pull/7138, so licensing is not an issue anymore.

bcbio/structural/__init__.py
naumenko-sa: A good point. bcbio's persistence, in general, is common knowledge among users, but I don't think it was documented. I have clarified it here: https://bcbio-nextgen.readthedocs.io/en/latest/contents/configuration.html#persistence

PR #3366, opened 2020-11-04
Drop hardcoded hg38 requirement in PureCN

lbeltrame:
PureCN can and will work fine also with hg19.

### DIFF
--- a/bcbio/structural/__init__.py
+++ b/bcbio/structural/__init__.py
@@ -136,23 +136,29 @@ def create_cnv_pon(samples):
     """create normal db using SNV PON and coverage information"""
     coverage_files = []
     sample0 = utils.to_single_data(samples[0])
+    genome = dd.get_genome_build(sample0)
     purecn_pon_build = tz.get_in(["config", "algorithm", "purecn_pon_build"], sample0)
     if purecn_pon_build:
+
         for sample in (utils.to_single_data(x) for x in samples):
             sample_coverage = tz.get_in(["depth", "bins", "purecn"], sample)
+
             if os.path.exists(sample_coverage):
                 coverage_files.append(sample_coverage)
+
         work_dir = tz.get_in(["dirs", "work"], sample0)
         out_dir = utils.safe_makedir(os.path.join(work_dir, "gemini"))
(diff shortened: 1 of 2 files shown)


bcbio/structural/__init__.py
```diff
@@ -136,23 +136,29 @@ def create_cnv_pon(samples):
...
                 coverage_files.append(sample_coverage)
+
         work_dir = tz.get_in(["dirs", "work"], sample0)
         out_dir = utils.safe_makedir(os.path.join(work_dir, "gemini"))
         purecn_coverage_files_txt = os.path.join(out_dir, "purecn_coverage_files.txt")
+
         with open(purecn_coverage_files_txt, "w") as out_handle:
-            for fname in coverage_files:
-                out_handle.write(fname + "\n")
+            out_handle.write("\n".join(coverage_files))
```
naumenko-sa, line 154: I think it is better to have an empty last line. I've met some tools that would break if you did not give them that last line.

bcbio/structural/__init__.py
```diff
@@ -136,23 +136,29 @@ def create_cnv_pon(samples):
...
+
         snv_pon = tz.get_in(["config", "algorithm", "purecn_snv_pon"], sample0)
-        normal_db = os.path.join(out_dir, "normalDB_hg38.rds")
+        normal_db = os.path.join(out_dir, "normalDB_{}.rds".format(genome))
```
naumenko-sa, line 157: thanks! I agree we should not hardcode the hg38 genome here. However, I have not tested hg19 purecn in bcbio. It would need some fixing - mappability resource as far as I remember. Could you help testing hg19?

naumenko-sa:
I'm merging this in bcbio, thanks so much for reviewing.
To make hg19 really working we would need a couple of recipes in cloudbiolinux:
https://github.com/chapmanb/cloudbiolinux/blob/master/ggd-recipes/hg38/purecn_mappability.yaml
and simple repeats.

bcbio/structural/__init__.py
lbeltrame: Indeed I can. I'll try to open a PR for cloudbiolinux.

lbeltrame:
Done, see https://github.com/chapmanb/cloudbiolinux/pull/372.

PR #3373, opened 2020-11-13
utils.py - python 3.8 -> 3.9 transition

smoe:


roryk:
Thanks so much for these fixes.

smoe:
@roryk I need to apologize for this one, OrderedDict is still in collections. For Python 2+3 (early+late) compatibility you may want to have this as
```
try:
   from collections.abc import Mapping
except ImportError:
   from collections import Mapping

from collections import OrderedDict
```
My correction came a few minutes to late for your merge. Sorry!!!

roryk:
No worries, thanks for the heads up!

PR #3374, opened 2020-11-14
heterogeneity/loh.h: SyntaxWarning: "is not" with a literal. Did you mean "!="?

smoe:
This warning popped up while testing the Debian package.


roryk:
Thank you!

Issue #3384, opened 2020-12-01
New Feature: create a tools_on: vardict_sv option


naumenko-sa:
Hello everyone!

By default, we are running vardict with --nosv. We need a yaml option to switch sv detecting in vardict **on**.
It should not be the default behavior, because the output could be problematic for some downstream analysis tools
(duplicated SV calls in the file?).

Switching SV detection ON adds 10-15% to vardict running time. Vardict itself is ~10% of a bcbio run, so it is not a big increase.

SN

lbeltrame:
+1 to the idea. Is there any paper (or preprint) which discussed VarDict's SV calling?

roryk:
Hi Luca, yup https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4914105/ talks about it some. We haven't done any independent validation though, yet.

naumenko-sa:
That request did not prove itself to be useful: it disturbs the default behaviour of vardict and does not add useful output.

PR #3392, opened 2020-12-10
add tools_on: vardict_sv option

naumenko-sa:
this is an experimental feature: https://github.com/bcbio/bcbio-nextgen/issues/3384
allows to specify:
```
tools_on:
  vardict_sv
```
and remove --nosv option from the vardict call.

It may cause some issues in the output vcf. 
Reported to brake 
```
tools_on:
  damage_filter
```

### DIFF
--- a/bcbio/variation/vardict.py
+++ b/bcbio/variation/vardict.py
@@ -50,10 +50,9 @@ def _vardict_options_from_config(items, config, out_file, target=None, is_rnaseq
     vardict_cl = get_vardict_command(items[0])
     version = programs.get_version_manifest(vardict_cl)
     # turn off structural variants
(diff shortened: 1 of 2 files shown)


lbeltrame:
Are the SVs included inline in the VCF among the other variants? If so, to be consistent with other tools doing SV calling, perhaps they should be moved to their own VCF?

naumenko-sa:
we are testing the output, not sure yet whether it is useful at all

roryk:
Do you think this is mergable now as an experimental option? Or leave it unmerged.

naumenko-sa:
let us leave it for now - the request is kind of frozen, it is not clear how useful vardict sv is, but some issues from it are already seen

PR #3429, opened 2021-01-28
Update error message in the heterogeneity caller check

kmendler:
## Problem
The error message that is given when a cnv caller is not specified for a heterogeneity caller does not include the name of the heterogeneity caller. This is a problem since TitanCNA (and previously PureCN) can raise this error, despite being given in the `svcaller` section of the yaml.

## Solution
The error message now lists the specified heterogeneity callers that also require a cnv caller.


roryk:
Thank you Kerrin!

PR #3439, opened 2021-02-18
add tools_off: contamination option

naumenko-sa:
since verifybamid2 hangs on AWS instances


roryk:
Great-- the QC modules are supposed to be toggleable via `tools_off` so this restores that behavior.

PR #3476, opened 2021-04-19
Update softclipping for chipseq

yoonsquared:
in chip-seq, for bowtie2, we changed '--very-sensitive' -> '--very-sensitive-local' to softclip.
This has been discussed with Rory and Meeta.
We can wait for Shannan to confirm this change.
Will update in slack for the pull-request.

### DIFF
--- a/bcbio/ngsalign/bowtie2.py
+++ b/bcbio/ngsalign/bowtie2.py
@@ -108,7 +108,7 @@ def filter_multimappers(align_file, data):
     bam.index(out_file, config)
     return out_file
 
-ANALYSIS = {"chip-seq": {"params": ["-X", 2000, "--very-sensitive"]},
+ANALYSIS = {"chip-seq": {"params": ["-X", 2000, "--very-sensitive-local"]},
             "variant2": {"params": ["-X", 2000]},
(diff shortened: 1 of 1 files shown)


lbeltrame:
A rationale for the change for those who are outside those communication channels would be very helpful.

yoonsquared:
I have confirmed with Shannan on moving forward with the soft-clipping as default.
We couldn't think of a case that soft-clipping will be detrimental to the aligning/peak-calling process.
- except in the cases where our reads become so short, and we cannot align them after soft-clipping (this is very rare).

In our real dataset that had the 5' contamination, we observed better mapping rate (91.9% -> 96.01%).
And as we offer BWA as the other option for ChIPseq data alignment, and bwa soft-clips, we want a more comparable steps between the two tools.

@naumenko-sa , Would it be possible to add a no-soft-clipping option?
We agreed on having the '--very-sensitive-local' as the default.
But have an option to take it out and use '--very-sensitive' instead.

Thank you!

Let us know if you have questions or concerns!

naumenko-sa:
Thanks for the explanation @yoonsquared !
Merging the default!

PR #3539, opened 2021-10-14
Mutect2 and PureCN: also genotype sites present in the PoN

lbeltrame:
This is actually recommended by the PureCN documentation, as otherwise
sites in the PoN, if present in more than two samples, will not get
genotyped by MuTect 2.

MuTect2 docs: https://gatk.broadinstitute.org/hc/en-us/articles/360037593851-Mutect2#--genotype-pon-sites
PureCN doc reference: https://www.bioconductor.org/packages/release/bioc/vignettes/PureCN/inst/doc/Quick.html#3_Create_VCF_files

### DIFF
--- a/bcbio/variation/mutect2.py
+++ b/bcbio/variation/mutect2.py
@@ -92,7 +92,7 @@ def mutect2_caller(align_bams, items, ref_file, assoc_files,
         broad_runner = broad.runner_from_config(items[0]["config"])
         gatk_type = broad_runner.gatk_type()
         # shared Mutect2 settings for PureCN analysis in the case of:
-        # - PON creation 
+        # - PON creation
         # - Tumor-only PureCN run
         # - T/N PureCN run
         # PURECN requirement alters Mutect2 variants calling!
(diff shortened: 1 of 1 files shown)


lima1:
@lbeltrame , @naumenko-sa . I don't have a lot of tumor/normal pairs. Do you have some comparison if pon sites flag is really necessary with --genotype-germline-sites? I wrote the documentation when Mutect2 introduced the features and never went back to test. Is it possible we just add artifacts?

naumenko-sa:
Hi Markus @lima1 !

I do have some tumor-normals (not a lot though and I have exomes only at the moment, not panels). 
Could you please clarify what the comparison would be?

1. Create a PON
2. Run TON+purecn with `mutecn -pon`
3. Run TON+Purecn without `mutect -pon`
4. Call variants with mutect as T/N
5. Compare variants called in 4, 3, 2
6. compare purecn calls in 2 vs 3

Sergey

lima1:
Hi Sergey @naumenko-sa ,

hm, don't think it needs to be that detailed for now. if you run mutect t/n with and without --genotype-pon-sites (both with --genotype-germline-sites and all other best practice recommendations) and look at the difference, are the added calls high quality germline or likely mostly noise with low allelic fraction?

I should have tried this long ago, but our pipelines are still at Mutect 1.1.7 for now and other things kept me busy.

Thanks so much,
Markus

naumenko-sa:
Hi Markus @lima1 !

I've ran 5 T/N WES pairs of some cancer samples with some ~200-300 somatic mutations and ~50k-60k calls overall per T sample and the results are identical for `--genotype-germline-sites true` and `--genotype-germline-sites true --genotype-pon-sites true`.

bcbio1.2.9/gatk4.2.4.0

SN

lima1:
Thanks a lot!

PR #3555, opened 2021-11-11
Adjust paired PureCN to the new way of locating PureCN.R

lbeltrame:
Unlike run_purecn_normaldb, run_purecn was never adjusted.


naumenko-sa:
Thanks for catching this! 
We never ran purecn with cnvkit or gatk segmentation in the last year or so - only with native + normaldb, that is why we missed it.

PR #3640, opened 2022-03-30
Update hg38-dbnsfp.conf

matthdsm:
Update conf for dbNSPF, since column names and numbers have changed since v3.5

PS, current version does not work or correspond to version in cloudbiolinux


matthdsm:
xref https://github.com/chapmanb/cloudbiolinux/pull/397

matthdsm:
cleaning up

PR #3673, opened 2022-09-13
Split up broad and narrow antibodies

Gammerdinger:
The previous version had an extra new line with broad and narrow on separate sides of the new line character and it was a bit confusing.


naumenko-sa:
thanks @Gammerdinger

Issue #3689, opened 2022-12-10
Error with bcbio_setup_genome.py: AttributeError: 'Namespace' object has no attribute 'cloudbiolinux'


amizeranschi:
Hi,

I'm getting an error when trying to create a custom genome. Here's the command I'm running and the error it produces:

```
$ bcbio_setup_genome.py -f GWHBDNW00000000.genome.fasta -g GWHBDNW00000000.gff --gff3 -i bwa seq -n GWHBDNW00000000 -b build1 --buildversion None
Traceback (most recent call last):
  File "/data/share/bcbio_nextgen/anaconda/bin/bcbio_setup_genome.py", line 249, in <module>
    cbl = get_cloudbiolinux(args, REMOTES)
  File "/data/share/bcbio_nextgen/anaconda/lib/python3.7/site-packages/bcbio/install.py", line 807, in get_cloudbiolinux
    cloudbiolinux_remote = remotes["cloudbiolinux"] % args.cloudbiolinux
AttributeError: 'Namespace' object has no attribute 'cloudbiolinux'

```

PR #3691, opened 2022-12-21
Add --cloudbiolinux argument

amizeranschi:
Fixes the following issue: https://github.com/bcbio/bcbio-nextgen/issues/3689

The problem originated from this commit: https://github.com/bcbio/bcbio-nextgen/commit/d61e77825f46548101db9b64776269f8e96ee220


amizeranschi:
Hi @naumenko-sa 

Have you had a chance to look at this?

naumenko-sa:
Thanks! I thought, we had it - but it was only in the main setup script, not setup genomes.

PR #3741, opened 2024-04-28
remove Python2 crumbs

a-detiste:


a-detiste:
ping

lpantano:
thank you! I think we would need to pump the version in this case. Have you tested the code with the test units?

a-detiste:
I don't remember... it's too far away and I have done hundreds such PR. I ll test again ASAP.

lpantano:
thanks, I was trying to add some test, but it is being difficult to get the env going. I will keep trying.

